PR #282, opened 2026-05-28
heddle#259: genericize Repository backends — dyn→generic + native async Pg (Phase 1)

### LINKED ISSUE #259: refactor(repo): genericize Repository backends — replace dyn dispatch with type parameters
## Motivation

`Repository` holds its three backends as trait objects:

```rust
store: Box<dyn ObjectStore>,
refs:  Box<dyn RefBackend>,
oplog: Box<dyn OpLogBackend>,
```

In practice the concrete type is always known at compile time — the CLI binary uses `FsStore + RefManager + OpLog`, the server uses `S3Store + PgRefBackend + PgOpLogBackend`. The `dyn` dispatch was an early simplification that now blocks making trait methods `async fn` (async methods aren't `dyn`-safe without boxing futures via `#[async_trait]`).

Converting to generics would let the Pg backends use native `async fn` in traits, drop the `RuntimeBridge::block()` shim on every call, and eliminate vtable overhead on the hottest paths (blob reads, ref lookups).

## Current state

| Trait | Implementors | `dyn` references | Async methods needed |
|---|---|---|---|
| `ObjectStore` | 6 (`FsStore`, `InMemoryStore`, `SharedStore`, `S3Store`, `DummyStore`, `CountingStore`) | ~72 | None today (S3 could benefit eventually) |
| `RefBackend` / `CoreRefBackend` | 2 (`RefManager`, `PgRefBackend`) | ~7 | Yes — `get_thread`, `get_marker`, `create_marker`, `resolve`; more methods could follow |
| `OpLogBackend` | 2 (`OpLog`, `PgOpLogBackend`) | ~14 | Yes — `recent_batches_scoped`, `undo_batches_scoped`, `redo_batches_scoped` + default wrappers |

### Phase 1 — `RefBackend` + `OpLogBackend` (high value, small blast radius)

1. Add type parameters to `Repository`: `Repository<R: RefBackend, O: OpLogBackend>` (keep `ObjectStore` as `dyn` for now).
2. Propagate the generics through `Repository`'s public API. Most consumers call `repo.refs()` or `repo.oplog()` and operate on the concrete return — the signature changes but the body doesn't.
3. Make all Pg-bound methods `async fn` in the traits. `RefManager` and `OpLog` impls return immediately (sync bodies in `async fn` are fine).
4. Drop `PgRefBackend::block()` / `PgOpLogBackend::block()` and the `RuntimeBridge` for methods that are now natively async.
5. Update `Importer`, `RefEmitter`, `OplogEmitter` to be generic over the backend instead of taking `&dyn`.
6. Update `resolve_refspec` to accept async closures (or make it async itself).

### Phase 2 — `ObjectStore` (larger blast radius, lower urgency)

`ObjectStore` has 6 implementors (including test doubles) and ~72 `dyn` references across proto, mount, repo, and tests. It also has no async methods today. Genericize only if/when S3 or a future backend needs async, or if the vtable cost shows up in profiles. Consider a type alias (`type LocalRepo = Repository<FsStore, RefManager, OpLog>`) to keep test ergonomics manageable.

## Key risks

- **Monomorphization bloat** — CLI and server each instantiate one concrete `Repository<…>`, so this is fine. Test code with `InMemoryStore` or mock backends adds one more instantiation per combination; monitor binary size.
- **Trait bound propagation** — Functions that are generic over `R: RefBackend` may need `Send + Sync + 'static` bounds depending on usage context. The `where` clauses will get noisy in places; type aliases help.
- **Partial async in sync contexts** — CLI commands run in a sync `main`. Callers that hold a `Repository<RefManager, OpLog>` never actually need to `.await` (the futures resolve immediately), but the compiler still requires an async context. Wrapping in `block_on` at the CLI entry point or using `pollster::block_on` for the trivially-ready futures are both options.

## Out of scope

- Making every method on every trait `async`. Only methods that the Pg backend actually awaits on need the `async` keyword. The rest stay sync.
- Removing `#[async_trait]` from other traits in the codebase — only the three backend traits are in scope.

## Definition of Done

- [ ] `Repository` is generic over `RefBackend` + `OpLogBackend`
- [ ] All Pg-backend methods that hit the database are `async fn` in the trait
- [ ] `PgRefBackend` and `PgOpLogBackend` no longer route through `RuntimeBridge::block_on` for async trait methods
- [ ] `cargo test` passes for both default features (CLI) and `--features postgres` (server)
- [ ] No `#[async_trait]` on `CoreRefBackend`, `RefBackend`, or `OpLogBackend`
---

## Scope decision (locked 2026-05-27)

Per the jj deep-dive at orchestrator root `jj-git-integration-deep-dive.md` §4.4 + Q4: this work is a **sealed-interface extraction**, not a public-extension-point design. Heddle remains the sole implementer; no formal pluggable mock backend; no FFI surface; trait methods may be private to heddle.

Rationale: pre-1.0 with no second-backend customer on the roadmap, the extraction work pays for itself purely by making `crates/cli/src/bridge/` testable in isolation. Designing for hypothetical future non-git backends multiplies the effort by ~3-4x and yields no current value.

Reference shape (template, not literal port): jj's `WorkingCopy` trait at `lib/src/working_copy.rs` — five methods (`name`, `workspace_name`, `operation_id`, `tree`, `sparse_patterns`, `start_mutation` returning `LockedWorkingCopy`). Note `tree()` returns a tree-id, not a worktree-state — the trait describes *which commit are we at*, not *what's on disk*. That split is right for heddle too.

Out of scope for this issue (kept for the future-pluggable case): formal pluggable backend, mock backend, full interface review for FFI stability.

thethorne48:
## Summary

Phase 1 of heddle#259: replace `dyn` backend dispatch on `Repository` with type parameters, and make the database-hitting Postgres methods native `async`.

- **`Repository<R: RefBackend = RefManager, O: OpLogBackend = OpLog>`.** Default type params keep the bare `Repository` name resolving to the local flavor, so the ~600 `&Repository` / `repo.refs()` / `repo.oplog()` call sites across the CLI are unchanged. `refs()`/`oplog()` now return concrete `&R`/`&O`; at heddle call sites `R = RefManager`/`O = OpLog`, so method calls resolve to the **inherent sync** methods — no async cascade through the sync CLI. `from_parts` is generic so the hosted server can build `Repository<…, PgRefBackend, PgOpLogBackend>`.
- **Native async backend reads (no `async_trait`).** The DB-hitting reads are now `-> impl Future + Send`: `CoreRefBackend::{get_thread, get_marker, create_marker, resolve}` and `OpLogBackend::{recent_batches_scoped, undo_batches_scoped, redo_batches_scoped}` + their wrappers. `PgRefBackend`/`PgOpLogBackend` `.await` `sqlx` directly for these — **`RuntimeBridge::block` dropped for them** (the remaining sync Pg methods keep the bridge, as the issue scopes). Spelled `impl Future + Send` rather than `async fn` so the futures carry an explicit `Send` bound (for the hosted Tower/tonic stack) and stay clean under `-D warnings` (`async_fn_in_trait`). Sealed interface — heddle is the sole implementer.
- **`Importer`/`RefEmitter`/`OplogEmitter` genericized** over the backend (`&dyn` is no longer object-safe once the traits have `async fn`). `import_git_into` stays sync and drives the always-ready local futures via `pollster::block_on` — safe even inside the CLI's Tokio runtime (pollster is not a Tokio runtime, so no nested-runtime panic).
- **Repairs the pre-existing `--features postgres` compile break.** `PgRefBackend` still used `&str`/`Vec<String>` after the `ThreadName`/`MarkerName` newtype migration (35 errors); `PgOpLogBackend` had two `'static`-escape errors. This path **is not gated by CI** (no workflow runs `--features postgres`), so it had silently rotted. Fixed here since the DoD requires postgres to pass and these are the files being rewritten. **Recommend adding `--features postgres` to `rust-tests.yml`** so this can't rot again.

Closes HeddleCo/heddle#259

## Phase 1 / Phase 2 split

Phase 2 (genericize `ObjectStore`) is **carved out to a follow-up** (paste-ready issue body posted as a comment on #259):

- **No async payoff** — `ObjectStore` has no `async` methods, so it delivers none of #259's motivation (native async / dropping `RuntimeBridge`).
- **Runtime store selection conflict** — `Repository::open` picks `FsStore` vs `S3Store` at *runtime* from `[storage.s3]` config (`build_store`). A compile-time `S` param can't express that without an `AnyStore` enum (re-introduces dynamic dispatch) or a `S = Box<dyn ObjectStore>` default (CLI keeps dyn dispatch — no perf win).
- **~72 dyn sites across 7 crates** would be a mega-diff with no async benefit.

The store therefore stays `Box<dyn ObjectStore>` this PR.

## Downstream coordination (weft)

weft consumes heddle via the **published** `heddle-repo = "0.2"` (no path/patch dep), so this branch does not build or break weft CI. When weft next bumps its heddle dependency it will need to update two call sites — `crates/weft-server/src/service/refs_hosted_impl.rs` and `.../derived_summaries/backfill.rs` — which call `Repository::from_parts(…, Box::new(PgRefBackend::new(…)), Box::new(PgOpLogBackend::new(…)), …)`: drop the `Box::new` (the generic `from_parts` takes `R`/`O` by value).

## Red commit

N/A — behavior-preserving type-system refactor (DoD Type ≠ TDD-Rust). The existing suite is the regression guard; the converted `ingest` tests now exercise the `async` backend reads end-to-end (`pollster::block_on(RefEmitter::emit(...))` awaits the async `get_marker`).

## Test evidence

Run in an **isolated target dir** (`CARGO_TARGET_DIR=…`), because the orchestrator shares one `target/` across worktrees and a concurrent worktree on an older branch was clobbering the shared `heddle-refs` artifact. CI uses a clean target, matching the isolated result.

```
cargo test --workspace
  → 77 test binaries, test result: ok — 3095 passed; 0 failed

cargo test -p heddle-refs  --features postgres  → ok. 58 passed; 0 failed
cargo test -p heddle-oplog --features postgres  → ok. 12 passed; 0 failed

bash scripts/check-no-silent-default-tree-load.sh   → clean (513 files scanned)
bash scripts/check-default-install-ships-worker.sh  → OK: heddle-fuse-worker built

cargo clippy --workspace --all-targets -- -D warnings                              → clean
cargo clippy -p heddle-refs -p heddle-oplog --features postgres --all-targets -- -D warnings → clean

cargo check -p heddle-cli --no-default-features --features git-overlay,semantic,zstd → ok
cargo check -p heddle-cli --no-default-features --features native,semantic,zstd       → ok (2 pre-existing dead-code warnings, unrelated)
```

## Manual verification

N/A — covered by tests (no UI surface).

---
<a href="https://app.blacksmith.sh/HeddleCo/codesmith/heddle/pr/282"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1782574412&installation_id=132405951&pr_number=282&repository=HeddleCo%2Fheddle&return_to=https%3A%2F%2Fgithub.com%2FHeddleCo%2Fheddle%2Fpull%2F282&signature=1e234b8693547a6c378c0de14a27efc6146a7eeffbe7a97c5741e56b4372c289"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>
<sup>Need help on this PR? Tag <code>@codesmith</code> with what you need. Autofix is disabled.</sup>


thethorne48:
## codecov/patch lift (r2)

Reproduced CI's coverage build locally (`cargo llvm-cov --locked --workspace --features git-overlay,native,semantic,zstd`) and intersected the lcov with the refactor diff. The 53.91% gap was **53 uncovered lines, none of them Pg code**:

| File | before | after |
|---|---|---|
| `oplog/oplog_backend.rs` (trait defaults) | 0/24 | 24/24 |
| `oplog/oplog_core.rs` (async overrides) | 9/13 | 13/13 |
| `refs/backend.rs` (default `resolve`) | 0/22 | 22/22 |
| `refs/refs_manager.rs` (async trait impls) | 1/4 | 4/4 |
| `repo/repository.rs`, `ingest/*` | already hit | — |
| **patch total** | **62/115 = 53.91%** | **115/115 = 100%** |

**(A) Coverable** — async trait methods the CLI never reaches because the concrete inherent (sync) methods take precedence. Covered by: trait-dispatch tests on the real `RefManager`/`OpLog`, plus two minimal in-memory backends (`MemRefBackend`, `MemOpLog`) that *keep* the trait defaults so the default `resolve` / `record_batch_scoped_if_no_transaction` bodies get exercised.

**(B) Pg-only** — `crates/refs/src/refs/pg_refs.rs` (75 added lines) and `crates/oplog/src/oplog/pg_oplog.rs` (46 added lines) are `#[cfg(feature="postgres")]`. The default-feature coverage build doesn't compile them, so llvm-cov emits **zero records** for those files and codecov never counts them in the patch denominator. **No `codecov.yml` ignore is required** — an ignore entry would be dead config. Real Pg coverage is heddle#284 (CI postgres service container).

Tests-only + `pollster` dev-deps on `heddle-refs`/`heddle-oplog`; no refactor-behavior changes. `cargo test --workspace` + `clippy --workspace --all-targets -D warnings` green.

thethorne48:
@codex please review the final state here. This is the Phase 1 generics refactor (#259): `Repository<R: RefBackend, O: OpLogBackend>` with `RefBackend`/`OpLogBackend` async-fn traits and generic `Importer`/`RefEmitter`/`OplogEmitter`. Behavior is unchanged from the pre-refactor `LocalRepo`; the recent commits only added tests (codecov/patch is now 100% on the diff). CI is green.

forge-zephyr:
@codex

thethorne48:
@codex please review the final state here. This is the Phase 1 generics refactor (#259): Repository<R: RefBackend, O: OpLogBackend> with RefBackend/OpLogBackend async-fn traits and generic Importer/RefEmitter/OplogEmitter. Behavior is unchanged from the pre-refactor LocalRepo; recent commits only added tests (codecov/patch is now 100% on the diff). CI is green.

PR #512, opened 2026-06-04
refactor(objects): pack install returns installed ids via one validated seam

### LINKED ISSUE #502: refactor(objects): make pack install the deep seam — streaming install should return installed ids
From the architecture-deepening review (candidate 6, "worth exploring", **verified**).

## Problem
Pack internals are deep but the install seam is shallow. `native_pack` (`crates/proto/src/native_pack.rs`) buffers pack/index data and tracks object ids **separately** because the streaming install path doesn't report installed ids — verified self-documenting: `install_pack_streaming` "Returns nothing" (`crates/objects/src/store/mod.rs:578`). Callers must know which path is memory-bounded vs id-returning vs validated vs cache-safe.

## Solution
Deepen the install path so the byte-buffer and temp-file installs share **one validated implementation** that produces the same installed-id result. Keep the public `ObjectStore` interface small. Clearer locality for checksum validation, id listing, cache invalidation, and memory-bounded native sync/import.

## AC
- [ ] Streaming + buffered pack install share one validated impl returning installed ids.
- [ ] `native_pack` no longer tracks ids separately; behavior identical.

thethorne48:
Closes #502.

## Problem
Pack internals were deep but the install seam was shallow. The byte-buffer install (`install_pack`) validated every entry and returned installed ids; the memory-bounded temp-file install (`install_pack_streaming`) did neither — it just `rename(2)`'d the staged files into place and "returned nothing". Callers (native sync) therefore tracked installed ids out-of-band, and the two paths diverged on validation and cache behavior.

## Change — one shared validated seam
- **`validate_and_list_pack`** (new, in `fs_impl.rs`): the shared validated core — checksum-validates every pack entry against its tagged id and returns the installed id list. Both install seams run their pack through it.
- **`install_pack`** (byte-buffer) now delegates validation/listing to the shared core and clears the recent-object caches, matching the streaming path's cache invalidation.
- **`install_pack_streaming`** now returns `Vec<PackObjectId>` and validates via an **mmap-backed `PackReader::open`** before renaming — same validation + same id set as the byte-buffer path, while keeping the pack out of the heap (memory-bounded promise intact). The reader's mmap is dropped before the rename.
- **Trait default `install_pack_streaming`** updated to return the ids it gets from `install_pack`; `AnyStore` dispatch return type updated to match.

## native_pack
`NativePackBundle` drops its separately-tracked `ids` field. `build_native_pack` no longer accumulates ids, and the install result is the single source of truth for installed ids (the receive path reads them off `install_pack`/`install_pack_streaming`). Stale "API can't report ids yet" note updated.

## Behavior
Identical objects installed, same validation, same cache state. The streaming path is *strengthened* (it now validates valid packs transparently). Public `ObjectStore` surface unchanged in shape — only the streaming method's return type.

## Test plan
- [x] `cargo test --locked --workspace --features git-overlay,native,semantic,zstd` — green (pack/store/native-sync tests are the oracle).
- [x] `cargo clippy --locked --workspace --all-targets --features git-overlay,native,semantic -- -D warnings` — clean.
- Updated native-sync receive test asserts `installed_ids == [PackObjectId::Hash(hash)]` from the install result instead of the dropped `bundle.ids`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---
<a href="https://app.blacksmith.sh/HeddleCo/codesmith/heddle/pr/512"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1783160638&installation_id=132405951&pr_number=512&repository=HeddleCo%2Fheddle&return_to=https%3A%2F%2Fgithub.com%2FHeddleCo%2Fheddle%2Fpull%2F512&signature=3e4508c5c06e91aec13f0ea065c9d6a57e150a48362342f586fab976ed00029d"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

### DIFF
--- a/crates/objects/src/store/fs/fs_impl.rs
+++ b/crates/objects/src/store/fs/fs_impl.rs
@@ -100,6 +100,24 @@ fn validate_action_serialized(data: &[u8], id: ActionId) -> Result<Action> {
     validate_loaded_action(&id, action)
 }
 
+/// Validate every entry in a pack against its tagged id (checksum
+/// validation) and return the installed id list. This is the shared
+/// validated core for both install seams: the byte-buffer install
+/// (`install_pack`) and the memory-bounded temp-file install
+/// (`install_pack_streaming`) both run their pack through here, so
+/// both apply the same checksum validation and report the same
+/// installed ids regardless of how the bytes reach the store.
+fn validate_and_list_pack(reader: &crate::store::pack::PackReader) -> Result<Vec<PackObjectId>> {
+    let ids = reader.list_ids();
+    for id in &ids {
+        let Some((obj_type, data)) = reader.get_object_bytes(id)? else {
+            continue;
+        };
+        validate_pack_entry(id, obj_type, data.as_ref())?;
+    }
+    Ok(ids)
+}
+
 fn validate_pack_entry(id: &PackObjectId, obj_type: ObjectType, data: &[u8]) -> Result<()> {
     match (id, obj_type) {
         (PackObjectId::Hash(hash), ObjectType::Blob) => validate_blob_bytes(data, *hash),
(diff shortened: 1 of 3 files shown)


thethorne48:
@codex please review — refactor(objects): the streaming pack-install path now returns installed ids through one shared validated seam (checksum validation + id listing + cache invalidation), so native_pack no longer buffers + tracks ids separately. Byte-buffer and temp-file installs go through the same validated core. Behavior identical (pack/store/native-sync tests are the oracle). Check & test green.

PR #513, opened 2026-06-04
refactor(cli): Cli::open_repo() + MUTATING spread + CommandRuntimeContract trim (part of #497)

thethorne48:
Part of #497 (CLI mechanical-dedup slice — does **NOT** close it; the semantic/mount/objects/glue slices remain for separate PRs).

## Summary
Three same-logic CLI dedups from the simplification backlog:

1. **`Cli::open_repo()`** — added one `impl Cli` method (lazy cwd resolution + `.context("open Heddle repository")`) and replaced **83** call sites: ~74 copy-pasted `Repository::open(cli.repo.as_ref().unwrap_or(&cwd))?` inlines across ~45 files + the 10 call sites of the 3 private `open_repo(cli)` helpers (`agent.rs`/`review.rs`/`conflict.rs`), whose definitions are deleted. The two non-canonical variants (`watch.rs` keeps its `.context("opening repository for watch")`; `status.rs` keeps `.ok()?`) route through the chokepoint too.
2. **`MUTATING` base contract** — rewritten as `CommandContract { mutates: true, supports_op_id: true, observe_only: false, may_move_ref: true, writes_heddle_refs: true, ..READ_JSON }`. Verified the other 30 fields match `READ_JSON` exactly, so the value is provably identical.
3. **`CommandRuntimeContract` dead-field trim** — dropped the never-read fields from the struct + `runtime_contract()` builder, keeping the 11 live ones (`path`, `display`, `supports_json`, `supports_op_id`, `persists_op_id`, `uses_bootstrap_op_id_store`, `json_kind`, `help_visibility`, `help_rank`, `surface`, `canonical_command`). The lockstep test `parsed_command_runtime_contract_exposes_catalog_fields` trimmed to match (kept `help_rank`). Every dropped field was confirmed read-free outside tests — all `.<field>` hits were on `CommandContract`/the catalog entry struct, not a live `CommandRuntimeContract`. **No fields skipped.**

Behavior identical; catalog JSON output reads the lightweight `CommandContract` directly and is unaffected by the runtime-contract trim. Net ~108 LoC removed.

## Test plan
- [x] `cargo test --locked --workspace --features git-overlay,native,semantic,zstd` — green (one doctest hiccup was shared-target contention; passes in isolation)
- [x] `cargo clippy --locked --workspace --all-targets --features git-overlay,native,semantic -- -D warnings` — clean
- [x] `heddle doctor docs --all` — no drift across 57 files

[...]

Issue #497, opened 2026-06-04
refactor: mechanical dedup backlog (cli/semantic/mount/objects/glue, ~-350 LoC)


thethorne48:
Low-risk, same-logic dedups found by the simplification sweep, grouped by area. Each is independent; pick off opportunistically when touching the area. Full detail: `.heddleco-orchestrator/reviews/simplification-backlog.md`.

### cli
- **`Cli::open_repo()`** — the idiom `Repository::open(cli.repo.as_ref().unwrap_or(&cwd))` is copy-pasted at **73 sites across 45 files** + 3 private `open_repo` helpers (conflict.rs/review.rs/agent.rs). One `impl Cli` method → single chokepoint + consistent error context. (-30 LoC, big consistency win)
- `MUTATING` catalog const re-spells all 35 fields; rewrite as `{ ...5 differing..., ..READ_JSON }` (command_catalog.rs:763). -30
- Trim ~26 never-read fields off `CommandRuntimeContract` (+ point the op-id/help wrappers at the `Copy` `CommandContract` instead of building the heavy struct). -50 (touches one test)
- `Cli::json(&repo)` shim for 134 `should_output_json(cli, Some(repo.config()))` sites. (char-count, not line-count)

### semantic
- `language_rules::classify_node` repeats function-like/container `Classified` construction ~12× → 2 helpers (like the existing `leaf_item`). -70
- `is_function_node` duplicates `is_function_kind`'s dispatch table → reimplement in terms of it. -23
- Delete vestigial `Language::parser_handle` pub pass-through (its "outside the crate" justification is false; 3 in-crate callers → `.parser()`). -6
- `analysis_similarity` Lines/Tokens branches identical but for tokenizer → one `jaccard()` helper. -15

### mount
- 4 read-side sites re-inline the existing `join_child` helper → call it. -12
- Hoist the triplicated `panic_payload_str` into the unconditionally-compiled `error.rs`. -16
- Collapse PendingHit→Entry construction (lookup + enumerate ×2) behind one converter. -55
- Share a `make_file_attr` builder between `file_attr_from`/`entry_attr_from` (14 identical fields). -15
- Drop the dead `#[cfg(not(target_os="linux"))]` branch in `fuse.rs` rename (module is already linux-gated). -7

### objects
- `versioned_msgpack_blob!` macro for the 6 container blobs (encode/decode/version-check boilerplate). -45
- Fold `Blob::from_slice` onto `From<&[u8]>`. -6

[...]

PR #514, opened 2026-06-04
refactor(mount/objects/glue): mechanical dedup (part of #497)

thethorne48:
Part of #497 (mount/objects/glue slices — does **NOT** close it; the semantic slice remains; the cli slice merged in #513).

Low-risk, same-logic dedups from the simplification sweep. Behavior is identical in every case — these are mechanical extractions, and the existing mount/objects/client/devtools test suites are the oracle.

## Items done

**mount (`crates/mount/src`)**
- 4 read-side empty-parent path joins (lookup/enumerate sites) → existing `join_child` helper.
- Triplicated `panic_payload_str` hoisted into the unconditionally-compiled `error.rs` as `pub(crate)` (gated to the union of shell backends so a no-shell build doesn't trip `dead_code`); fuse/fskit/projfs guard wrappers now call it.
- Duplicated `PendingHit`→`Entry` construction in `lookup` + `enumerate` collapsed behind one `entry_from_pending_hit` converter (intern/kind/size/mode derivation preserved exactly).
- Shared `make_file_attr(...)` builder between `file_attr_from`/`entry_attr_from` (14 identical fields).
- Dropped the dead `#[cfg(not(target_os="linux"))]` branch + redundant `#[cfg(target_os="linux")]` attrs in `fuse.rs` `rename` (module is already linux-gated).

**objects (`crates/objects/src`)**
- `versioned_msgpack_blob!` macro emitting `FORMAT_VERSION`/`new`/`encode`/`decode`/version-checked `validate` for the **5 uniform** container blobs: structured_conflict, discussion, risk_signal, state_review, state_context. Each keeps only its per-item `validate()`. Error variant + version preserved per type (state_context uses `InvalidEncoding` + version 2).

**glue (`crates/client`, `crates/devtools`)**
- `authed_call!` macro for the ~21 authenticated `grpc_hosted/user.rs` user-RPC dispatches (apply_auth → await → `map_err(status_to_protocol_error)` → `into_inner`), collapsing each call site.
- Shared `devtools::asserter` module (`for_each_rs_file` + `read_search_dirs`/`read_allowlist` env-readers); the 4 `check_*.rs` binaries now supply only their skip predicate + visitor + env-key constants. Skip runs before the read so `files_scanned` semantics are unchanged.
- `to_proto_object_info` delegates to `object_descriptor_with_status(info, Present, "")`.

## Items skipped (reported)

- **objects: fold `Blob::from_slice` onto `From<&[u8]>` / drop inherent `from_slice`.** Skipped. Dropping the inherent method forces `&[u8]`-exact arguments at ~16 call sites across 5 crates, several of which pass `&Vec<u8>` or `&[u8; N]` literals (`b"..."`) that rely on deref/unsizing coercion the generic `From<&[u8]>` impl can't provide. Converting them all adds `.as_slice()`/`[..]` noise and risks breaking test compiles across crates — net-negative churn for a -6 item, not a clean mechanical fold.
- **objects: redaction not folded into the macro.** `RedactionsBlob` diverges from the other 5 — its `decode` does no version check, it uses a separate `Decoding` error variant (no `UnsupportedVersion`), and it carries extra methods (`empty`/`push`/`has_active`/`latest`/`mark_all_purged`). Forcing it into the uniform macro would change behavior, so it's left as-is.

## Test plan

- [x] `cargo test --locked --workspace --features git-overlay,native,semantic,zstd` — green.
- [x] `cargo clippy --locked --workspace --all-targets --features git-overlay,native,semantic -- -D warnings` — clean.
- [x] `cargo clippy --locked -p heddle-mount --features fuse --all-targets -- -D warnings` — clean (exercises the OS-gated fuse shell edits).
- [x] `bash scripts/check-no-silent-default-tree-load.sh` — clean (538 files scanned).

Net **−195 LoC** (566 insertions / 761 deletions).

---
<a href="https://app.blacksmith.sh/HeddleCo/codesmith/heddle/pr/514"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1783166101&installation_id=132405951&pr_number=514&repository=HeddleCo%2Fheddle&return_to=https%3A%2F%2Fgithub.com%2FHeddleCo%2Fheddle%2Fpull%2F514&signature=187ee6b197c469f8a7c92643c88b8d49289130764fda700879671c5414e74031"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

### DIFF
--- a/crates/cli/src/cli/cli_args/cli_base.rs
+++ b/crates/cli/src/cli/cli_args/cli_base.rs
@@ -83,6 +83,24 @@ pub struct Cli {
     pub op_id: Option<String>,
 }
 
+impl Cli {
+    /// Open the Heddle repository the command should act on: the `--repo`
+    /// path if given, otherwise the current working directory (resolved
+    /// lazily so a supplied `--repo` never touches the cwd).
+    pub fn open_repo(&self) -> anyhow::Result<repo::Repository> {
+        use anyhow::Context as _;
+        let cwd;
+        let repo_path = match self.repo.as_ref() {
+            Some(path) => path,
+            None => {
+                cwd = std::env::current_dir().context("get current working directory")?;
+                &cwd
+            }
+        };
+        repo::Repository::open(repo_path).context("open Heddle repository")
+    }
+}
+
 #[derive(Copy, Clone, Debug, ValueEnum)]
 pub enum OutputMode {
     Json,
--- a/crates/cli/src/cli/commands/actor_cmd.rs
+++ b/crates/cli/src/cli/commands/actor_cmd.rs
@@ -272,7 +272,7 @@ pub async fn cmd_actor_spawn(
     if let Some(name) = thread.as_deref() {
         ThreadId::new(name).map_err(|err| anyhow!(thread_name_invalid_advice(&err)))?;
     }
-    let repo = Repository::open(cli.repo.as_ref().unwrap_or(&std::env::current_dir()?))?;
+    let repo = cli.open_repo()?;
 
     let base_state = repo.head()?.ok_or_else(|| {
         anyhow!(RecoveryAdvice::repository_no_head_capture_first(
@@ -421,7 +421,7 @@ pub async fn cmd_actor_spawn(
 }
 
 pub async fn cmd_actor_list(cli: &Cli, active_only: bool) -> Result<()> {
-    let repo = Repository::open(cli.repo.as_ref().unwrap_or(&std::env::current_dir()?))?;
+    let repo = cli.open_repo()?;
     let registry = AgentRegistry::new(repo.heddle_dir());
 
     let mut entries = registry.list()?;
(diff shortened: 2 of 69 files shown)


thethorne48:
@codex please review — refactor [part of #497]: mechanical dedup across mount (join_child reuse, panic_payload_str hoisted to error.rs, PendingHit→Entry converter, make_file_attr builder, dead fuse cfg branch), objects (versioned_msgpack_blob! macro for the 6 container blobs, Blob::from_slice fold), and glue (one authenticated-RPC dispatch helper for the ~23 user.rs sites, shared for_each_rs_file in devtools, to_proto_object_info delegate). -215 LoC; behavior identical (existing tests are the oracle). Check & test green.

Issue #549, opened 2026-06-06
fix(release): Linux binaries link glibc 2.39 — crash on Debian 12 / Ubuntu 22.04 (pin an older floor)


thethorne48:
Surfaced by the #328 apt-hosting spike (`docs/design/apt-hosting-gpg-spike.md` §cross-cutting risk), verified against `release.yml`.

## The bug (live, affects current releases)
`release.yml` builds the two Linux legs on **glibc 2.39** runners:
- `x86_64-unknown-linux-gnu` → `ubuntu-24.04` (release.yml:196-197)
- `aarch64-unknown-linux-gnu` → `ubuntu-24.04-arm` (release.yml:193-194)

A binary dynamically linked against glibc 2.39 fails with `version 'GLIBC_2.3x' not found` at runtime on older glibc:
- **Debian 12 (bookworm)** — glibc 2.36
- **Ubuntu 22.04 LTS** — glibc 2.35

So today's Linux release tarballs **install/extract cleanly and then crash on first run** on two of the most common LTS targets. This affects **all** Linux binary distribution channels, not just apt:
- the raw GitHub-Release tarballs users download now,
- the future Homebrew Linux block (#232),
- the future apt channel (#234) — which is why #328 flagged it as a hard prerequisite.

## Fix — pin the Linux release build to an older glibc floor
Pick one (impl decides; tradeoffs noted):
- **Build the gnu legs on `ubuntu-22.04` / `ubuntu-22.04-arm`** (glibc 2.35). Simplest; a 2.35-linked binary runs on Debian 12 (2.36) and Ubuntu 22.04+ forward. Lowest-risk, smallest diff.
- **`cargo-zigbuild`** targeting an explicit glibc version (e.g. `*-unknown-linux-gnu.2.31`) from any runner. Most control over the floor; adds a build dep.
- **musl static target** (`*-unknown-linux-musl`): no glibc dependency at all, one portable static binary. Maximum portability; evaluate musl's DNS/perf tradeoffs for heddle's workload before committing.

## Acceptance criteria
- [ ] Linux release binaries run on Debian 12 + Ubuntu 22.04 (and newer) — verified, not assumed (run the built binary in a 2.35/2.36 container).
- [ ] `RELEASING.md` documents the supported glibc floor / target so it's a declared contract.
- [ ] The pipeline-contract asserter (`scripts/check-release-pipeline.sh`) covers the chosen floor (runner pin or target triple) so it can't silently regress to a newer runner.
[...]

PR #1050, opened 2026-07-18
feat(discuss): wire hosted discussion sync into the CLI

thethorne48:
## What
Wires the heddle CLI's `discuss` surface to the hosted weft **CollaborationService** so discussions authored on one clone are visible on another. Previously `heddle discuss` wrote local-only state-attachments that never reached the server (the #549 pack scan deliberately rejects client-sent attachments; no CLI verb published them). Push-time sync + clone/pull fetch.

- New `crates/client/src/grpc_hosted/collaboration.rs` (hosted CollaborationService client: OpenDiscussion/AppendTurn/…).
- New `crates/cli/src/client/discussion_sync.rs` (sync orchestration), wired into `push`, `clone`, `pull`, and `remote`.

## Verified end-to-end (fresh docker weft, merged main, enforce mode)
```
# author on clone proj:
heddle discuss open main.rs main 'Body needed?'; heddle discuss append <id> 'Print hello.'
heddle push weft  → [ok] synced 1 discussion(s) to dtester/proj

# fresh clone CL:
heddle clone …/dtester/proj CL
heddle discuss show <id>  → renders BOTH turns (Body needed? / Print hello.)

# bidirectional: append on CL → push → pull on proj:
heddle push origin  → [ok] synced 1 discussion(s)
heddle pull weft    → [ok] synced 1 discussion(s) from dtester/proj  (proj now shows CL's reply)
```

## Known minor issues (follow-up, non-blocking)
- A pulled turn can display **duplicated** (dedup on pull-merge).
- `repository has no HEAD` error syncing discussions on a repo pushed without a proper `main` HEAD (e.g. git-default `master`).
- Discussion local-ID differs across clones (content/anchor/turns round-trip correctly).

Implemented by a background agent; **round-trip verified by me**. Fable review requested (correctness + perf). Context/review sync are the same seam, follow-up. Do not self-merge.

---
<a href="https://app.blacksmith.sh/HeddleCo/codesmith/heddle/pr/1050"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1786980283&installation_id=132405951&pr_number=1050&repository=HeddleCo%2Fheddle&return_to=https%3A%2F%2Fgithub.com%2FHeddleCo%2Fheddle%2Fpull%2F1050&signature=48451fb992586cc228d4d9fc9b5351552ba13fcca421b0e9394e0c988f7e830e"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

### DIFF
--- a/crates/cli/src/cli/commands/clone.rs
+++ b/crates/cli/src/cli/commands/clone.rs
@@ -1623,6 +1623,18 @@ async fn clone_network(
                 .context("failed to materialize hosted clone worktree")?;
         }
         configure_hosted_clone_origin(&local_repo, &endpoint_spec, repo_path)?;
+        // Read path for hosted discussions (heddle discuss): materialize the
+        // hosted CollaborationService discussions for the cloned head into the
+        // local op-log so `discuss list` / `discuss show` see them. Best-effort:
+        // a fetch hiccup warns rather than failing an otherwise-good clone.
+        match crate::client::discussion_sync::pull_discussions(&local_repo, &mut client, repo_path)
+            .await
+        {
+            Ok(_) => {}
+            Err(error) => {
+                eprintln!("{} discussion sync skipped: {error:#}", style::warn_marker());
+            }
+        }
         if should_output_json(cli, Some(local_repo.config())) {
             let output = heddle_clone_output(
                 origin_url.clone(),
--- a/crates/cli/src/cli/commands/remote/mod.rs
+++ b/crates/cli/src/cli/commands/remote/mod.rs
@@ -1257,6 +1257,31 @@ async fn push_network(repo: &Repository, options: PushNetworkOptions<'_>) -> Res
     .await?;
     clear_line(&progress);
 
+    // Write path for hosted discussions (heddle discuss): the pushed state(s)
+    // are now on the server, so replay any local symbol-anchored discussions to
+    // the hosted CollaborationService over the same signed client. Best-effort:
+    // the object push already succeeded, so a discussion-sync hiccup warns
+    // rather than failing the push (the next push resumes from the mirror map).
+    if result.success {
+        match crate::client::discussion_sync::push_discussions(repo, &mut client, &repo_path).await
+        {
+            Ok(count) if count > 0 && !should_output_json(options.cli, Some(repo.config())) => {
+                println!(
+                    "{} synced {count} discussion(s) to {}",
+                    style::ok_marker(),
+                    style::dim(&repo_path)
+                );
+            }
+            Ok(_) => {}
+            Err(error) => {
+                eprintln!(
+                    "{} discussion sync skipped: {error:#}",
+                    style::warn_marker()
+                );
+            }
+        }
+    }
+
     // CLI maps wire/protobuf transport fields → pure domain fields; core
     // parses success/failure and builds execution facts / outcome.
     let fields = HostedPushResultFields {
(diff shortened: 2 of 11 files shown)


thethorne48:
**Fable review: BLOCK.** The happy-path round-trip works, but the sync model has correctness/provenance bugs under ordinary bidirectional use:

- **B1 — no turn identity:** `synced_turns` is a single ordinal count treated as a common-prefix on both sides, but local (timestamp) order and server (append) order diverge once both sides append. Result: a push can re-publish a *pulled peer turn* under the pusher's authenticated identity (misattribution), silently drop the pusher's own turn, and mint durable server-side duplicates that accumulate. Fix: per-turn identity — client-side `(author, posted_at, body)` match as a first cut (server turn-ids are the real fix), AND filter the push candidate set to *locally-authored* unsynced ops (don't re-push pulled turns).
- **B2 — mirror persisted only at loop end + abort-on-first-error:** a mid-loop failure leaves op-log writes done but no mirror saved → next pull re-materializes everything (unbounded local dup), and next push re-opens each as a new server discussion (misattribution). Fix: save mirror per-discussion + on the error path; collect-and-continue instead of `?`.
- **No-HEAD / head-advance (weft-side, deeper):** weft `latest_state_attachment` is strictly per-state with no carry-forward, so after any head-advancing content push, `AppendTurn` 404s and `ListByState(head)` returns empty — sync silently dies for pre-existing discussions. Caps the feature at single-state repos until fixed weft-side. Filing as a prerequisite.

Signing/enforce and the pull single-RPC shape are good. Sending back for B1+B2.

thethorne48:
**Fable re-review: BLOCK (round 2).** The per-turn link-set architecture is the right shape and B2 is genuinely fixed — but the two identity assumptions each have a hole that reproduces the B1 class (silent loss + misattribution):

- **F1 (blocking) — body-only reconciliation false-matches across authors** (`discussion_sync.rs:488-507`). On the normal pull-before-push path, `available` includes the user's OWN unpushed turns, and the match is body-equality alone. Two same-body turns from different authors ("lgtm", "+1") → A's local turn links to B's server turn → B's turn never materializes on A AND A's turn is marked linked so it's never pushed. Silent, permanent. **Fix:** require exact AUTHOR equality in reconciliation — only re-link turns we pushed (self + our hosted principal) or previously pulled (exact author match); anything else materializes as new, never body-link.
- **F2 (blocking) — `CollabOpId` not unique for LegacyImported discussions** (`materialize.rs:228-231` emits N root turns sharing one `root_id`). Append key collides → turn 3 Conflicts → mirror links root_id → retry drops turns 3..N of exactly the migrated repos this PR touches. **Fix:** turn identity = `(op_id, turn_index_within_op)` throughout (TurnLink, candidates, append key), or explicitly skip LegacyImported-rooted discussions with a visible message.
- **F3 (fix or accept) — self-filter fail-open:** `author_is_self` returns true on `self_attr=None` → with mirror loss + missing attribution, foreign turns get republished; and env-vs-config principal drift silently misclassifies own turns as foreign. **Fix:** return false on uncertainty (skip+warn); warn when skipping unlinked non-self turns.

Non-blocking (follow-ups): replay is per-bearer-token not per-subject; ordinal-as-identity only stable by accident (→ server turn ids); mirror-loss dup at discussion granularity; perf nit (fsync mirror even when unchanged). The weft#638 single-state cap stays an accepted documented limitation.

PR #1053, opened 2026-07-18
feat(context,review): wire hosted sync for context annotations + review signatures

thethorne48:
## What
Completes the CLI hosted-collaboration surface: wires **context annotations** and **review signatures** to hosted sync, the same seam `discuss` got in #1050 (`context` and `review` were local-only).

- New `crates/cli/src/client/context_sync.rs` (802L) + `review_sync.rs` (317L) + hosted `state_review.rs` client, reusing #1050's hard-won correctness model: per-item identity (not ordinal counts), **author-aware reconciliation** (never body-alone → no cross-author misattribution), fail-closed self-filter, per-item mirror save + collect-and-continue, graceful weft#638 head-advance degradation.

## Verified
- **Context forward sync (live, docker weft, enforce):** `context set` → `push` → `[ok] synced 1 annotation(s)` → fresh clone `context list` shows it. ✅
- **Unit tests:** context_sync 4/4 (incl `identical_bodies_distinct_ids_materialize_separately` — the cross-author/identical-body case that broke discussions), review_sync tests pass.

## Limitations (inherited, tracked)
Same weft#638 (per-state attachment, no carry-forward → single-state-repo cap) + weft#640 (no server item ids → same-user-identical edge). `review sign` remains the low-level primitive (external signature) — sync is wired; ergonomic `review approve` is a separate UX follow-up.

Follow-up to #1050. Do not self-merge (Fable review + verification follow).

---
<a href="https://app.blacksmith.sh/HeddleCo/codesmith/heddle/pr/1053"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1786997200&installation_id=132405951&pr_number=1053&repository=HeddleCo%2Fheddle&return_to=https%3A%2F%2Fgithub.com%2FHeddleCo%2Fheddle%2Fpull%2F1053&signature=e9457675c54865f194981dd9c9726d149b42bf347f7470e8ed2414725cda497a"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

### DIFF
--- a/crates/cli/src/cli/commands/clone.rs
+++ b/crates/cli/src/cli/commands/clone.rs
@@ -1635,6 +1635,15 @@ async fn clone_network(
                 eprintln!("{} discussion sync skipped: {error:#}", style::warn_marker());
             }
         }
+        // Read path for hosted context annotations (heddle context): materialize
+        // the hosted head's annotations into the local Context attachment so
+        // `context list` sees them. Best-effort, mirroring discussions.
+        match crate::client::context_sync::pull_context(&local_repo, &mut client, repo_path).await {
+            Ok(_) => {}
+            Err(error) => {
+                eprintln!("{} context sync skipped: {error:#}", style::warn_marker());
+            }
+        }
         if should_output_json(cli, Some(local_repo.config())) {
             let output = heddle_clone_output(
                 origin_url.clone(),
--- a/crates/cli/src/cli/commands/mod.rs
+++ b/crates/cli/src/cli/commands/mod.rs
@@ -18,7 +18,7 @@ mod command_catalog;
 mod commit;
 pub(crate) mod compact;
 mod completion;
-mod context;
+pub(crate) mod context;
 mod daemon;
 mod diff;
 mod discuss;
--- a/crates/cli/src/cli/commands/remote/mod.rs
+++ b/crates/cli/src/cli/commands/remote/mod.rs
@@ -1280,6 +1280,40 @@ async fn push_network(repo: &Repository, options: PushNetworkOptions<'_>) -> Res
                 );
             }
         }
+
+        // Write path for hosted context annotations (heddle context) and review
+        // signatures (heddle review sign) — same seam as discussions. #549
+        // rejects these attachments in the pack, so they only reach the server
+        // over the caller-authenticated RPCs. Best-effort: the object push
+        // already landed, so a sync hiccup warns rather than failing the push.
+        match crate::client::context_sync::push_context(repo, &mut client, &repo_path).await {
+            Ok(count) if count > 0 && !should_output_json(options.cli, Some(repo.config())) => {
+                println!(
+                    "{} synced {count} annotation(s) to {}",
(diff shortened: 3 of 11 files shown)


thethorne48:
**Fable review: BLOCK.** Six findings, same silent-loss classes #1050 fixed for discussions + one review-seam killer. Summary:

- **P1-1 (review sync non-functional):** weft `SignState` rejects `|signed_at − now| > 300s` (state_review.rs:48,619), so any signature minted >5min before push fails forever + warns every push (failures aren't mirror-recorded, state stays in the 50-scan). **Resolution:** relax the server's signed_at gate for PoP-authenticated hosted-sync installs — the server already runs `ensure_signing_key_belongs_to_caller`, so a historical `signed_at` from an authenticated caller is legitimate for a durable review-signature install (the gate is a replay-guard for the interactive mint, not for sync). Weft-side change to SignState. Plus record permanent failures in the mirror so a genuinely-rejected sig stops retrying.
- **P1-2 (context revision silent loss):** `synced_revisions: usize` prefix-count → concurrent two-author revisions silently drop one + duplicate the other. Fix: reconcile by `revision_id` set-difference in server order (proto `ContextRevision.revision_id` exists); return the minted revision_id from `ReviseContextResponse` (or author+content match for self).
- **P1-3 (supersede chain lost):** `supersedes_annotation_id` is a LOCAL id checked against SERVER ids (context_sync.rs:437-457) → own supersede degrades to create → both stay live on other clones. Fix: resolve local→server via the mirror before the check.
- **P2-4 (recovery body-alone, forbidden class):** minted-id recovery (context_sync.rs:477-485) matches content-alone → can adopt another author's annotation. Fix: require `candidate.attribution == authenticated_username()` (like discussion_sync rule i).
- **P2-5 (lost-mirror permanent wedge):** stable op-id + no nonce → dedup Replay → 'could not recover minted id' warn forever. Fix: author-aware adopt + per-attempt nonce written to mirror before the RPC.
- **P2-6 (review self-filter wrong source):** gates on `config().principal` but sign stamps via `get_principal()` (env→config→git) → own sigs silently never sync in git-overlay/env. Fix: use `get_principal()` + warn when unresolvable.

Also: the 'adversarial' unit tests (`identical_bodies_distinct_ids...`, `actor_self_filter...`) don't exercise the real reconcile/recovery path — replace with tests that do. Sending back.

Issue #1100, opened 2026-07-25
Split heddle-cli (100k LOC) + heddle-objects into 7 layered crates — audited design & phased plan


thethorne48:
# Workspace-crate split design for HeddleCo/heddle

**Basis:** `origin/main` @ `fc1d2623` ("refactor(refs): route ref writes through the AtomicMutation
chokepoint (#1098)"), worktree clean and fast-forwarded. All LOC/gate/edge numbers below are measured
on that tree, not estimated.

**Method, and how it differs from the weft design.** Every edge claim here was extracted from source
with comments and string literals **stripped first**, then cross-checked against the crate's declared
Cargo dependencies. Both edge kinds were collected: internal `crate::…` edges *and* external
`use <dep>::…` edges. That mattered immediately — a naive grep reported `crates/objects/src/object/`
depending on `state_review`, `repo`, `semantic` and `wire`, none of which are in `objects`'
`Cargo.toml`. All four were doc-links in `///` comments. Same class of artifact produced the phantom
`cli_args → commands::repo::ThreadMode` edge (a doc-link at `commands_args.rs:654`, pointing at a
module that does not exist — stale, worth fixing separately). This is the weft §2 mistake; it is
excluded here by construction.

---

## 0. Verdict summary

Heddle is **already a 22-crate workspace**, so this is not one 125k-LOC monolith like `weft-server` —
it is four fat crates inside an existing DAG. The verdicts differ per crate, and two of them are
"don't":

| Crate | LOC (src / tests) | Verdict | One-line reason |
|---|---|---|---|
| `cli` (`heddle-cli`) | 100,823 / 76,057 | **Split — 5 new crates** | Real ownership seams (args, render, verb contract, harness); commands are free functions, so movable |
| `objects` (`heddle-objects`) | 31,042 / 25 | **Split — 2 new crates** | Clean model/pack/store layering; one 5-reference cycle to break |
| `repo` (`heddle-repo`) | 58,014 / 479 | **Deferred — out of scope** | 27,268 LOC across 33 files is welded to `Repository` by inherent impls. The two clean leaf clusters (9,047 LOC) are a viable later track; the welded core is a separate design. See §5.3. |
| `core` (`heddle-core`) | 39,911 / 0 | **Do not split** | Cohesive plan layer, ~60 small independent files, no tangle to relieve, and an existing devtools gate hardcodes `crates/core/` as the sole owner of verification definitions |

[...]

Issue #271, opened 2026-05-27
fix(output): remove OutputMode::Auto — eliminate JSON-when-piped surprise


thethorne48:
## Motivation

`OutputMode::Auto` (in `crates/cli/src/cli/cli_args/cli_base.rs:92-97`) is still a real variant. When `cli.output == Some(Auto)`, the resolver hits `OutputFormat::Auto` → `!is_tty()` → emits JSON if piped.

Round 3 #256 marked this as fixed; round 4 marked it verify-and-likely-resolved. **Round 5 confirms it is NOT fixed.** PR251 fixed the default surface (default=text, regardless of pipe state) but left the `Auto` mode behavior intact. Help text at `crates/cli/src/cli/help.rs:125` still mentions ``--output auto` renders text on a TTY and JSON when piped``.

Effects:
- **Regular AI dev (R5)**: running `heddle status > /tmp/log` with default config produces JSON in the log, not text — surprising.
- **Agent (A1)**: subprocess.run() invocations under default config get JSON on pipe vs TEXT on TTY; agents defensively re-issue with `--output json`.

## Plan

1. Remove the `OutputMode::Auto` variant.
2. Remove the `--output auto` mention from `help.rs:125`.
3. Ensure `OutputFormat::Auto` config still routes to text (or rename the underlying enum to make this explicit).
4. Test: piped invocation with NO `--output` flag → text (post-PR251 default — regression-protect).
5. Test: piped invocation with `--output json` → JSON.
6. Lint: grep for `OutputMode::Auto` returns zero hits.

## Persona context

Round 5 finding R5 / A1 — cross-persona (all three).

## Closes

#256

## Reference

`persona-eval-round-5.md` at orchestrator root.

## Definition of Done

- [ ] `OutputMode::Auto` variant removed
- [ ] `help.rs` no longer mentions `--output auto`
- [ ] CLI test: piped invocation, no flag → text
- [ ] CLI test: piped invocation, `--output json` → JSON
- [ ] `cargo test --workspace` + clippy green

thethorne48:
[...]

PR #1102, opened 2026-07-25
refactor(cli): extract heddle-cli-args crate (phase A1)

thethorne48:
Phase **A1** of the crate-split tracked in #1100 (phase A0 landed as #1101).

Extracts the CLI's clap argument surface into a new workspace crate `heddle-cli-args` (`crates/cli-args/`).

> **Numbering:** this was "A2" in the original design. Track A was re-ordered by [erratum 2](https://github.com/HeddleCo/heddle/issues/1100#issuecomment-5080390827) after `heddle-cli-render` turned out not to be a leaf.

## What moved

All of `crates/cli/src/cli/cli_args/` — 21 files, ~4,513 LOC — internal paths mirrored (`src/cli/cli_args/…`). Diff shows `crates/{cli => cli-args}/…` rename detection; `git log --follow` traverses the move.

## Edges — audited before the move

Verified with an extractor catching `use super::`, `use crate::` **and** external `use <dep>::` together (the original design missed the first two classes — that is what erratum 2 is about). Complete outward set was 8 sites: 4 `crate::cli::{Cli, Commands, …}` self-references through the facade path (those types are *defined* in `cli_args`) inside `#[cfg(test)]` modules, plus `api::…::ReviewKind`, `objects::object::VisibilityTier`, and two `cli_shared::OutputMode`. **No upward edge into `commands`.**

## Call-site churn: zero

`crates/cli/src/cli/mod.rs` had a 90-line explicit `pub use cli_args::{…}` block plus three feature-gated re-export groups. That collapses to:

```rust
pub use heddle_cli_args as cli_args;
pub use heddle_cli_args::*;
```

No `use` line changed anywhere else in the workspace.

## One relocation worth flagging

`user_config_or_exit`'s `OnceLock<UserConfig>` cache moved into `heddle-cli-args` as `Cli::user_config_or_exit()`, because `cli_base.rs` (in the move set) needs it. Semantics identical — same `OnceLock`, same `load_default().unwrap_or_default()`, still one singleton per process — and the comment recording the #271 error-envelope bug stayed with the public wrapper in `cli/mod.rs`, which now delegates. Slight scope creep beyond a pure move; flagged for review.

## Features

`git-overlay` (11), `client` (5), `ingest` (4), `semantic` (4) forwarded from `heddle-cli` to `heddle-cli-args`. **None deleted** — per design §6, no gate in this effort becomes a crate boundary.

## Verification — all re-run independently of the implementing agent

| Gate | Result |
|---|---|
| `cargo build --locked --workspace` (default features) | pass |
| slim OSS build (`--no-default-features --features git-overlay,native,semantic,zstd`) | pass |
| `cargo clippy --workspace --all-targets -- -D warnings` | pass |
| `cargo test --workspace` | 1 pre-existing env failure, see below |
| `scripts/check-facade-render-free.sh` | pass |
| `scripts/check-cli-dep-count.sh` | pass — 360, ceiling 360, **unchanged** (workspace members aren't counted; erratum 1) |
| `scripts/check-no-silent-default-tree-load.sh` | pass, 651 files |

**Ratchets unchanged**, as required for a phase that doesn't touch `src/cli/commands`: `render_lint` 656 violations / 103 files; `typed_error_lint` 155 sites / 103 files. No baseline moved.

The 4 warnings in the slim OSS build (`remote_ops.rs:273`, `mount_lifecycle.rs:68/1006`, `harness/mod.rs:2295`) are **pre-existing** — verified identical on the A0 base.

### The one test failure is environmental, not a regression

`multi_agent_worktrees::e2e::agent_capture_and_ready_require_authenticated_writer_authority` fails locally with `anthropic/gpt-5.3-codex` where `openai/gpt-5.3-codex` is expected.

**Root cause traced:** `harness_policy.rs:122` derives `program` from **argv[0]**, and `detect_harness_kind` matches `program.contains("claude")`. The dev worktree lives at `…/heddle/.claude/worktrees/…`, so argv[0] of the binary under test contains `claude` → `HarnessKind::ClaudeCode` → `default_provider() = "anthropic"`, which takes precedence over the test's explicit `HEDDLE_AGENT_PROVIDER=openai` (see the comment at `harness_policy.rs:134`).

Confirmed **pre-existing** — fails identically on the A0 base, whose CI was fully green, because CI checks out to a path without `claude`. Not touched here, but the substring match on a full argv[0] path is arguably over-broad and may deserve its own issue.

---
<a href="https://app.blacksmith.sh/HeddleCo/codesmith/heddle/pr/1102"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1787604486&installation_model_id=21489&pr_number=1102&repository=HeddleCo%2Fheddle&return_to=https%3A%2F%2Fgithub.com%2FHeddleCo%2Fheddle%2Fpull%2F1102&signature=893aa53a677a710ffa2e90d7458423ba8791592b7f3214f8d2b82d3543cedc24"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


thethorne48:
**Fixed the CI failure in `0e0c5e24`.**

`crates/cli/tests/tier_coverage.rs` reads `commands_main.rs` through chained `.join()` calls:

```rust
PathBuf::from(env!("CARGO_MANIFEST_DIR")).join("src").join("cli").join("cli_args").join("commands_main.rs")
```

This phase moved `cli_args/` to `crates/cli-args/`, so that path stopped existing and the test panicked with `No such file or directory`. Repointed at the sibling crate.

**Why the original verification missed it — worth recording, because it is a verification-method bug, not a one-off.**

`cargo test --workspace` aborts at the first failing test *binary*. The known-environmental `multi_agent_worktrees` failure (see the PR description — argv[0] contains `claude` in this dev worktree) halted the run **before `tier_coverage` ever executed**. The output read `1 failed`, which I took as a complete result. It was cargo stopping.

So the accepted-as-environmental failure was actively masking a real one. A known-failing test in the suite silently caps what fail-fast verification can observe.

**Corrected method, now standard for every remaining phase of #1100:** `cargo test --workspace --no-fail-fast`, and count binaries rather than grep for failures. Re-run on this branch:

```
test binaries run : 91
binaries FAILED   : 1     <- multi_agent_worktrees only (environmental, pre-existing)
```

The earlier run never got past a handful of binaries.

**Design gap this exposes:** §8's path-keyed gate register listed `tier_coverage.rs` under the *contract* phase, because I profiled it as reading a commands file. It reads a **`cli_args`** file, so it belonged in this phase's gate list and was not in the brief. Recorded as an erratum on #1100.

thethorne48:
## Review: approve, ready to merge

The only PR in the stack that targets `main`, and therefore **the only one whose green checkmark means what it appears to mean** — 12 checks including `check-and-test`, `postgres-tests`, the three mount smokes, and `facade-render-free`. See HeddleCo/heddle#1100 (comment) for why that matters for the rest of the stack.

**On the relocation you flagged.** Moving `user_config_or_exit`'s `OnceLock<UserConfig>` into `heddle-cli-args` as `Cli::user_config_or_exit()` is the right call and the flag was worth raising. It is a genuine behaviour-preserving move — same `OnceLock`, same `load_default().unwrap_or_default()`, one singleton per process — and keeping the #271 error-envelope comment with the public wrapper that now delegates preserves the reason someone would need it. Scope creep in the strict sense, but the alternative (leaving the cache above a crate that needs it) would have forced a worse edge.

**Zero call-site churn is the strongest evidence in this PR.** Collapsing a 90-line explicit `pub use` block plus three feature-gated groups down to two lines, with no `use` line changing anywhere else in the workspace, is what a true move looks like. A split that required touching call sites would mean the boundary was in the wrong place.

Good catch on erratum 2 — the original edge analysis grepping `crate::` only, and missing `use super::`, is the same class of bug #1108 found in the design's verification command.

PR #1103, opened 2026-07-25
refactor(cli): extract heddle-cli-render crate (phase A2)

thethorne48:
Phase **A2** of the crate-split tracked in #1100. **Stacked on #1102 (phase A1) — merge that first**; this PR will retarget to `main` automatically.

Extracts the CLI's terminal-presentation layer into `heddle-cli-render` (`crates/cli-render/`).

## What moved

`style.rs`, `render.rs`, `render/{fsck,query}.rs`, `progress_render.rs`, `tips.rs` — ~1,433 LOC, `git mv` with internal `src/cli/…` paths preserved. `git log --follow` on `render.rs` traverses back through `d8701f2e`, `f362c886`, `8978ce54`.

## Not the leaf the design claimed

The original design called this a dependency-free leaf and scheduled it *before* `cli-args`. Both were wrong — see [erratum 2](https://github.com/HeddleCo/heddle/issues/1100#issuecomment-5080390827). §2.2's reverse-edge analysis grepped `crate::` only, so it missed `use super::cli_args::Cli` (`style.rs:31`) and the external `use repo::{Repository, Thread, ThreadManager}` / `use objects::…` / `use heddle_core::{FsckReport, QueryReport}` edges.

Actual dependency set: `heddle-cli-args`, `repo`, `objects`, `heddle-core`, `anstyle`, `anstream`, `blake3`, `hex`. Track A was re-ordered so args lands first. **`heddle-cli-render` does not depend on `heddle-cli`** — the layering is still strictly downward.

## The `should_output_json` upward edge

`render.rs` and `progress_render.rs` imported `crate::cli::should_output_json`, defined in `cli/mod.rs`, which stays in the facade. Resolved by the pure/glue rule: the resolver moved **down** to `heddle_cli_args::should_output_json` (`cli_base.rs:130`), next to the `Cli`/`OutputMode` types it inspects, and the facade re-exports it. Signatures and call sites unchanged.

## Visibility promotions

Seven items became `pub` so facade command code can still reach moved progress internals: `COMMIT_TICK_INTERVAL`, `format_transfer_bytes`, `progress_for`, `TerminalSink`, `TerminalSink::new`, `clear_line`, `finish_line`. `style::force_for_test` also crosses the boundary but stays compiled only for render-crate tests or the private `test-utils` feature. Clippy required a `Default` impl for the now-public `TerminalSink::new`.

This is the first phase where the weft "mirror the root and visibility stays invariant" prediction did **not** fully hold — mirroring kept the *paths* stable but these seven were genuinely reached across the new boundary.

## Path-keyed gates

- `crates/cli-render` added to `[workspace] members`.
- `scripts/check-facade-render-free.sh`: `heddle-cli-render` added to the **forbidden** pattern, so domain crates cannot depend on it either. Its own manifest correctly *not* added to the checked-manifest list (it legitimately owns `anstyle`/`anstream`).

| Gate | Result |
|---|---|
| `cargo build --locked --workspace` (default features) | pass |
| slim OSS build | pass |
| `cargo clippy --workspace --all-targets -- -D warnings` | pass, 0 errors |
| `cargo test --workspace --no-fail-fast` | **93 binaries, 1 failed** (environmental — see below) |
| `scripts/check-facade-render-free.sh` | pass |
| `scripts/check-cli-dep-count.sh` | pass — 360, ceiling 360, unchanged |
| `scripts/check-no-silent-default-tree-load.sh` | pass, 653 files |

**Ratchets unchanged** (this phase does not touch `src/cli/commands`): `render_lint` 656 / 103 files; `typed_error_lint` 155 / 103 files. No baseline moved.

The single failure is `multi_agent_worktrees::e2e::agent_capture_and_ready_require_authenticated_writer_authority`, pre-existing and environmental — `harness_policy.rs:122` reads `program` from argv[0] and matches `program.contains("claude")`, and this dev worktree lives under `.claude/worktrees/`. CI checks out to a clean path. Detail in #1102.

> Note on method: verification uses `--no-fail-fast` and counts test binaries. Plain `cargo test --workspace` halts at the first failing binary, and the environmental failure above was masking a real one in #1102.

---
<a href="https://app.blacksmith.sh/HeddleCo/codesmith/heddle/pr/1103"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1787608570&installation_model_id=21489&pr_number=1103&repository=HeddleCo%2Fheddle&return_to=https%3A%2F%2Fgithub.com%2FHeddleCo%2Fheddle%2Fpull%2F1103&signature=c8fac5ff6923679b5c0069ddf41afdb5511fbb7fb13e642e03588f1c1b1e23bd"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


thethorne48:
## Review: approve on substance — but this PR has not been tested

⚠️ **The Rust suite never ran here.** `rust-tests.yml` (plus `audit`, `unwrap-inventory`, `release-pipeline-check`) is gated `on: pull_request: branches: [main]`, and this targets `refactor/crate-split-a1-cli-args`. It got three checks — `verify`, `pr-to-status`, `add-issue-to-project` — where #1102 got twelve. The green checkmark here means something different from #1102's, and nothing in the UI says so.

Retargeting after #1102 merges will **not** fix it: base changes fire `pull_request: edited`, which is not among the default activity types. Force a `synchronize` (empty commit) after retarget and wait for the full suite before merging. Structural fix proposed in HeddleCo/heddle#1100 (comment).

**The "environmental" failure is properly diagnosed, and I checked it.** `harness_policy.rs:122` reads `program` from `argv[0]` and branches on `program.contains("claude")`, so a worktree under `.claude/worktrees/` silently gets different test policy. That is a real explanation with a named mechanism, not a hand-wave — and the underlying harness behaviour is a bug in its own right, since test policy should not depend on the filesystem path of the checkout. Filing that separately.

**The `--no-fail-fast` methodology note deserves more credit than it got.** Discovering that plain `cargo test --workspace` halts at the first failing binary and was therefore *masking a real failure in #1102* is the kind of thing that silently invalidates a whole verification table. Worth making it the default for this effort.

**The seven visibility promotions are fine**, and the honesty about the weft prediction not holding is useful signal for the rest of the split — mirroring the crate root preserved *paths*, but heddle's facade reaches across boundaries more than weft's `server`/`service` split did. Expect that to worsen in A3, and it does.

Adding `heddle-cli-render` to the *forbidden* pattern in `check-facade-render-free.sh` while correctly leaving it out of the checked-manifest list (it legitimately owns `anstyle`/`anstream`) is the right shape — the gate stays meaningful instead of being widened to accommodate the new crate.

PR #1104, opened 2026-07-25
refactor(cli): extract heddle-cli-contract crate (phase A3)

thethorne48:
Phase **A3** of the crate-split tracked in #1100 — the largest and highest-risk phase. **Stacked on #1103 (A2), which is stacked on #1102 (A1).** Merge in order; bases retarget automatically.

Extracts the CLI verb-contract surface into `heddle-cli-contract` (`crates/cli-contract/`): **17,606 LOC**, the only phase that removes source from `crates/cli/src/cli/commands/`.

## What moved

`command_catalog/`, `schemas.rs`, `doctor_docs.rs`, `doctor_schemas.rs`, `advice.rs`, `verification_health/`, `help.rs` — `git mv`, internal paths mirrored. `git log --follow` on `command_catalog/mod.rs` traverses back through `8fe4984c`, `09114772`.

The set is **wider than the original design** — `advice.rs` and `verification_health/` were pulled in because `schemas.rs:24` and `doctor_schemas.rs:36` import `super::{advice::RecoveryAdvice, verification_health::{…}}` while `verification_health/mod.rs:32` imports `command_catalog` back. That is a cycle; the original 5-unit scope could not compile. See [erratum 2](https://github.com/HeddleCo/heddle/issues/1100#issuecomment-5080390827); the maintainer approved the widening.

## ⚠️ 89 forced `pub(crate)` → `pub` promotions

The design (§10 risk 5) predicted this would be near-zero, on weft's evidence that mirroring the crate root keeps every visibility boundary meaning the same thing. **That did not transfer.** Breakdown:

| Area | Sites | What |
|---|---:|---|
| `advice` | 50 | `DIRTY_WORKTREE_CAPTURE_COMMAND`, facade-used `RecoveryAdvice` constructors/adapters |
| `verification_health` | 31 | re-exports, result types, profiles, preflight constructors, policy/adaptation fns |
| `command_catalog` | 8 | `ActionFields` + constructors, `heddle_action`, action-template parse/validate helpers |

Mirroring preserved *paths*, but heddle's contract surface is reached into by facade command code far more pervasively than weft's `server`/`service` boundary was. **This is the main thing to review** — it is a real widening of the public API surface, and worth checking whether some of these should instead be `#[doc(hidden)]` or re-exported more narrowly.

## ⚠️ Ratchets lowered to exactly the measured actual

Per the confirmed ratchet policy (never raise; re-measure downward when a scan root shrinks):

| Ratchet | Baseline | Actual | Files |
|---|---|---|---|
| `RENDER_VIOLATION_BASELINE` | 840 → **647** | 656 → 647 | 103 → 95 |
| `MAX_UNTYPED_ANYHOW_SITES` | 157 → **154** | 155 → 154 | 103 → 95 |

**Note this leaves zero headroom** — both gates now sit at exactly their measured value, so any newly-introduced violation fails CI immediately. That is a deliberate tightening (the old 840-vs-656 gap was 184 violations of unclaimed slack, flagged as a latent hazard when A0 recorded the pre-split values), but it is a change in the gate's character from "don't regress much" to "don't regress at all". Accept deliberately or say so and I'll restore proportional slack. The `MIN_SCANNED_RUST_FILES` floors from A0 are unchanged at 50.

## Feature gates: zero deleted

Before/after counts match exactly — client 14, git-overlay 15, ingest 4, semantic 3. Per design §6, no heddle feature becomes a crate boundary in this effort.

One manifest subtlety: a standalone `cli-contract` with `default = []` failed seven moved tests, so the crate defaults to its four contract features while `heddle-cli` depends on it with `default-features = false` and forwards explicitly.

## Path-keyed gates — all seven, plus three the design missed

Updated: `advice_lint.rs` (now scans both facade and contract roots, mirrored allowlist keys preserved), `check-authored-states-signed.sh` (scans both command trees — security-relevant, would otherwise have silently stopped covering the moved code), the moved `doctor_docs` source walker (scans both trees), both ratchets, workspace membership, `tier_coverage`'s stale diagnostic path, and stale source-path references in docs/generated code.

Verified needing no change: `op_id_coverage` (reads `main.rs`), `ts_types_in_sync` (generated artifacts), `agent_api_schema` (snapshot), ref-write exemptions (name no moved file), mount-lane selector (not an objects-track crate).

**The brief was wrong** about tier/op-id/TS/agent gates needing repointing — none read the moved files. Recorded as errata.

| Gate | Result |
|---|---|
| `cargo build --locked --workspace` (default) | pass |
| slim OSS build | pass (4 pre-existing warnings, identical on base) |
| `cargo clippy --workspace --all-targets -- -D warnings` | pass |
| `cargo test --workspace --no-fail-fast` | **95 binaries, 1 failed** (environmental) |
| `cargo test -p heddle-mount` | pass |
| `check-authored-states-signed.sh` | pass, both trees |
| `check-default-cli-contracts.sh` | pass, 157 verbs, no drift |
| `check-cli-dep-count.sh` | pass — 360, ceiling 360, unchanged |
| `check-facade-render-free.sh` | pass |
| `check-no-silent-default-tree-load.sh` | pass, 656 files |

**Correction to the implementing agent's report:** it stated the `multi_agent_worktrees` harness test "passed this time" and reported 0 failing binaries. It does not pass — re-running it directly still yields `anthropic/gpt-5.3-codex`. The failure is the known environmental one (`harness_policy.rs:122` matches `program.contains("claude")` against argv[0]; this worktree lives under `.claude/worktrees/`). Binary count (95) and ratchet numbers were independently confirmed correct.

`cargo fmt --all --check` reports pre-existing drift in unrelated CLI client files; confirmed **not a CI gate** in `.github/workflows/`.

---
<a href="https://app.blacksmith.sh/HeddleCo/codesmith/heddle/pr/1104"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1787611126&installation_model_id=21489&pr_number=1104&repository=HeddleCo%2Fheddle&return_to=https%3A%2F%2Fgithub.com%2FHeddleCo%2Fheddle%2Fpull%2F1104&signature=2368d2b8097d47ee5cf042028ca457c306a0927cd73c700d26f0e4b53b65bc7d"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


thethorne48:
## Review: answering your two flagged questions — and one blocker

### ⚠️ Blocker: the ratchet changes in this PR were never verified by CI

This PR lowers two ratchets to **exactly** their measured values:

| Ratchet | Baseline | New |
|---|---|---|
| `RENDER_VIOLATION_BASELINE` | 840 | 647 |
| `MAX_UNTYPED_ANYHOW_SITES` | 157 | 154 |

**The gates that enforce those numbers did not run on this PR.** `rust-tests.yml` and `unwrap-inventory.yml` are gated `on: pull_request: branches: [main]`, and this targets the A2 branch — so it got three checks, not twelve. Those values are currently a local measurement only.

At zero headroom that matters more than usual: off by one in either direction gives you a permanently red `main` or a gate that is silently looser than advertised. Force a `synchronize` after retargeting to `main` and confirm both gates pass in CI before merge. Detail in HeddleCo/heddle#1100 (comment) — note that retargeting alone will not trigger a run.

### Your question 1: zero headroom — keep it

You asked whether to accept zero headroom deliberately or restore proportional slack. **Keep zero.**

The old 840-vs-656 gap was 184 violations of unclaimed slack, and a gate sitting well above the thing it guards is the specific failure this org has already paid for — a bundle-size ratchet set above the value a real leak produces, which therefore could not fire on the regression it existed to catch. "Don't regress at all" is the correct character for a ratchet; "don't regress much" is a gate you are guessing about.

The change is right. It just needs the numbers confirmed by the gate itself rather than by a local run.

### Your question 2: the 89 `pub` promotions — lower risk than you flagged

You called this "the main thing to review". It is less severe than it looks, for a reason worth recording: `PUBLISHABLE_CRATES` in `publish-crates.yml` is `heddle-format, heddle-objects, heddle-crypto, heddle-runtime-bridge, heddle-schema, heddle-oplog`. **`heddle-cli-contract` is not published**, so these 89 are not a crates.io semver commitment. The cost is internal API clarity, not compatibility, and `#[doc(hidden)]` would buy little.

Accept them for this phase. The useful follow-up is measurable rather than aesthetic: count which of the 89 the facade actually references across the boundary versus which were promoted defensively during the move, and narrow the second group. That is a mechanical audit, not a judgement call, and it belongs after the split rather than inside it.

### On the widened scope

Pulling in `advice.rs` and `verification_health/` was forced by a genuine cycle (`schemas.rs:24` and `doctor_schemas.rs:36` importing `super::{advice, verification_health}` while `verification_health/mod.rs:32` imports `command_catalog` back) — the original 5-unit scope could not have compiled. Right call, and documenting it as an erratum rather than quietly widening is the right way to handle it.

The manifest detail — `cli-contract` defaulting to its four contract features while `heddle-cli` depends on it with `default-features = false` and forwards explicitly — is worth keeping in the phase notes. That is exactly the kind of thing that looks arbitrary in six months.

Issue #836, opened 2026-07-01
discuss: discussion becomes unrecoverable via `discuss show` after the state advances (does not follow HEAD, not flagged orphaned)


thethorne48:
## Summary

A discussion opened with `heddle discuss open` is anchored to the current state. Once the state advances (even by an unrelated change), the discussion becomes effectively unreachable through the primary commands, and is not reported as orphaned. The data is preserved, but only discoverable if you already know the original anchor state id.

This contradicts the `heddle discuss` help text, which states: "Anchors travel across renames and cross-file moves on subsequent state mutations."

## Repro

```
# in an adopted / git-overlay repo at HEAD state S0
heddle discuss open src/Surface.zig draw "review question..."   # -> discussion hd-DISC, anchored to S0
heddle discuss list                                             # -> shows hd-DISC (lists discussions on HEAD)
heddle discuss show hd-DISC                                     # -> shows the discussion

# advance the state with any change (does not need to touch Surface.zig)
heddle context set --path src/App.zig --scope symbol:tick --kind rationale -m "..."
heddle checkpoint -m "advance state"                            # HEAD now S1

heddle discuss show hd-DISC          # Error: Some requested entity was not found: discussion hd-DISC not found
heddle discuss list                  # (no discussions)          -- HEAD-default hides it
heddle discuss list --status orphaned# (no discussions)          -- not flagged orphaned either
heddle discuss list --state S0       # still there, status open   -- only way to find it
```

## Expected

After a state advance, at least one of:
- the discussion follows HEAD forward (per the help text), or
- `heddle discuss show <id>` resolves it regardless of HEAD (or accepts a `--state`), and/or
- it is reported under `discuss list --status orphaned`.

## Actual

`discuss show <id>` fails permanently (no `--state` option on `show`), the discussion silently drops off `discuss list` (which defaults to HEAD), and it is not classified `orphaned`. It is only recoverable via `discuss list --state <original-anchor>`, which requires knowing that id.

## Impact

The reflection itself is fine (the discussion clones/pulls through weft), but after any subsequent commit the review thread is undiscoverable by id in normal use. This also affects the original author's repo, not just clones.

## Environment

- heddle 0.6.0 (installed release build) and a client-enabled 0.6.0 debug build
- git-overlay repo (adopted), hosted remote `heddle://127.0.0.1:8421/...` (weft dev)

PR #1129, opened 2026-07-28
Harden release pipelines and repository security controls

thethorne48:
## Summary

- Harden crate publishing and release workflows with validation checks.
- Add repository redaction, purge, signing, and state-visibility controls.
- Improve remote target handling, clone planning, hosted sync, and CLI completion.
- Add regression coverage for redaction, signing, and production security features.
- Update changelog and release documentation.

## Testing

- Added and updated CLI and production-feature integration tests.
- Added publish and release pipeline validation scripts.
- Verified the affected crates with Clippy under `-D warnings`.
- Ran all 413 `heddle-core` library tests and the signed-purge propagation regression.

## Compatibility and rollout notes

- Redaction signing payloads intentionally make a clean break from the unused `hd-redact-v1` format. Existing v1 signatures do not verify under v2. Any future payload-version change that needs compatibility must add explicit verify-side version dispatch and operator-facing migration errors.
- Purge lifecycle transitions are re-signed by the active client identity. Receivers fail closed unless that client public key is trusted.
- Automated trusted-client-key provisioning through the Weft/Tapestry control plane is not implemented by this PR. Hosted purge synchronization must remain disabled until that provisioning is delivered; current local synchronization requires explicit `heddle redact trust add` enrollment.

---
<a href="https://app.blacksmith.sh/HeddleCo/codesmith/heddle/pr/1129"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1787799708&installation_model_id=21489&pr_number=1129&repository=HeddleCo%2Fheddle&return_to=https%3A%2F%2Fgithub.com%2FHeddleCo%2Fheddle%2Fpull%2F1129&signature=ef2c7c2a772a38b43cbfcc52ab27fac7b0af2fca8d726a080cadabe720ad4d33"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/.github/workflows/publish-crates.yml
+++ b/.github/workflows/publish-crates.yml
@@ -121,7 +121,7 @@ jobs:
       # job entirely on no-op merges.
       has_publishes: ${{ steps.classify.outputs.has_publishes }}
     steps:
-      - uses: actions/checkout@v7
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7
         with:
           # Full history isn't needed (we don't walk it), but we DO
           # need the commit to be a real commit object so we can
@@ -294,17 +294,9 @@ jobs:
     if: needs.validate-publish.outputs.has_publishes == 'true'
     runs-on: ubuntu-24.04
     timeout-minutes: 60
-    env:
-      # The env var name (CARGO_REGISTRY_TOKEN) is cargo's documented
-      # name and is what `cargo publish` reads at runtime — it must
-      # stay exactly this. The GitHub Actions secret name
-      # (CRATES_IO_API_KEY) is independent: it's the label under repo
-      # Settings → Secrets and variables → Actions. The strict asserter
-      # verifies both halves of this mapping — any rename on either
-      # side would silently break authentication.
-      CARGO_REGISTRY_TOKEN: ${{ secrets.CRATES_IO_API_KEY }}
+    environment: release
     steps:
-      - uses: actions/checkout@v7
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7
         with:
           # Pin to the SHA validate-publish verified, NOT
           # refs/heads/main. main is mutable; a force-push between
@@ -313,35 +305,27 @@ jobs:
           # as release.yml.
           ref: ${{ needs.validate-publish.outputs.commit_sha }}
 
-      - uses: dtolnay/rust-toolchain@stable
+      - uses: dtolnay/rust-toolchain@4cda84d5c5c54efe2404f9d843567869ab1699d4 # stable
 
-      - uses: Swatinem/rust-cache@v2
+      - uses: Swatinem/rust-cache@e18b497796c12c097a38f9edb9d0641fb99eee32 # v2
         with:
           shared-key: publish-crates
           cache-on-failure: true
 
-      - name: Assert CARGO_REGISTRY_TOKEN is present
-        shell: bash
-        run: |
-          set -euo pipefail
-          # We deliberately check at the FIRST credentialed step (not
-          # in validate-publish) so the validation pass still runs and
-          # surfaces what *would have* been published, even if the
-          # secret isn't configured yet. That gives the operator a
-          # punch list to act on. If the validation pass shipped
-          # without a publish attempt, it'd be silent.
-          if [[ -z "${CARGO_REGISTRY_TOKEN:-}" ]]; then
-            echo "::error::CARGO_REGISTRY_TOKEN env var is empty — the CRATES_IO_API_KEY secret is unset or the mapping in publish-crates.yml is wrong. Add CRATES_IO_API_KEY under repo Settings → Secrets and variables → Actions, then re-run."
-            exit 1
-          fi
-          echo "ok: CARGO_REGISTRY_TOKEN is present (length=${#CARGO_REGISTRY_TOKEN})"
-
       - name: Publish each crate (idempotent, retry on 5xx)
         shell: bash
         env:
+          # Keep the crates.io credential scoped to the only step that
+          # needs it. Checkout, toolchain, and cache actions never receive it.
+          CARGO_REGISTRY_TOKEN: ${{ secrets.CRATES_IO_API_KEY }}
           TO_PUBLISH: ${{ needs.validate-publish.outputs.to_publish }}
         run: |
           set -euo pipefail
+
+          if [[ -z "${CARGO_REGISTRY_TOKEN:-}" ]]; then
+            echo "::error::CARGO_REGISTRY_TOKEN env var is empty — the CRATES_IO_API_KEY secret is unset or the mapping in publish-crates.yml is wrong. Add CRATES_IO_API_KEY to the approval-protected release environment, then re-run."
+            exit 1
+          fi
 
           # Pretty-print the publish plan up front so the workflow log
           # has a clean record before any side effects.
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -27,9 +27,11 @@ name: Release binaries
 # the same FSKit SDK floor.
 #
 # Signing: cosign keyless via the Sigstore public-good instance. No
-# stored secrets — the GitHub OIDC token is the trust anchor. Verifying
-# parties run `cosign verify-blob --certificate-identity-regexp ...`;
-# see RELEASING.md for the exact recipe.
+# stored secrets — the GitHub OIDC token is the trust anchor. Every
+# signing or publishing job is gated by the approval-protected `release`
+# environment before GitHub issues credentials. Verifying parties run
+# `cosign verify-blob --certificate-identity-regexp ...`; see
+# RELEASING.md for the exact recipe.
 
 on:
   push:
@@ -41,17 +43,16 @@ on:
   workflow_dispatch:
     inputs:
       tag:
-        description: 'Existing tag to build (vX.Y.Z or vX.Y.Z-(rc|alpha|beta)[.N]); or a synthetic prerelease tag when branch_dry_run is true.'
+        description: 'Existing tag to build (vX.Y.Z or vX.Y.Z-(rc|alpha|beta)[.N]).'
         required: true
       branch_dry_run:
-        description: 'TEMP: build from the selected branch/ref without publishing a GitHub Release.'
+        description: 'Deprecated and refused: release credentials are never exposed to branch-selected workflow code.'
         type: boolean
         required: false
         default: false
 
 permissions:
-  contents: write   # create release + upload assets
-  id-token: write   # cosign keyless OIDC
+  contents: read
 
 concurrency:
   group: release-${{ github.ref }}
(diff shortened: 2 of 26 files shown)


thethorne48 (commented):
Reviewed the full diff. The security substance here is good, and a few pieces are better than good. Notes below, then two things to fix and one to confirm.

## What's right

**`purged_at` in the signed payload** closes a real hole. The lifecycle field previously sat outside the signature, so a relay could forge a destructive purge transition. Covering it is the correct fix and the added assertion in `mark_purged_is_idempotent_and_observable` pins it.

**The signature-eclipse fix is correct, and the test is better than most.** Moving from `latest_state_attachment` to "consider every signature attachment, valid if any verifies" uses the right quantifier: forging a valid signature requires the private key, so extra attachments cannot manufacture validity — but under the old code an attacker who appended a *corrupted* attachment with a later `created_at` made a legitimately-signed state report `Invalid`. Pairing it with the append-only rejection on superseding covers the other half.

What makes this land is that the test doesn't stop at the `Repository` API — it installs `packed_sibling` with `supersedes: None` directly through `repo.store()`, simulating a received pack landing below the acceptance seam. That's the bypass that would otherwise have been missed.

**Release-pipeline hardening is textbook least privilege:**
- actions SHA-pinned rather than tag-pinned,
- `CARGO_REGISTRY_TOKEN` narrowed from job-level env to the single step that publishes, so checkout/toolchain/cache never see it,
- an approval-gated `release` environment in front of every credentialed job,
- top-level `contents: read`, with `id-token: write` re-declared only on the signing jobs and `contents: write` only on the job that creates the release.

I checked that last point specifically, because dropping top-level `id-token: write` silently breaks cosign keyless signing if the jobs don't re-declare it. They do.

**Removing `branch_dry_run` is the strongest change in the PR.** A branch-selected workflow definition being able to request Apple credentials and an OIDC token is a direct privilege-escalation path — the workflow can rewrite the validation that precedes the credential request. Keeping the input around to fail existing callers explicitly, rather than deleting it outright, is the right migration.

## Fix: the version-tag doc comment is now false

`REDACTION_SIGNING_PAYLOAD_VERSION_TAG` moves `hd-redact-v1` → `hd-redact-v2`, and nothing retains v1 — there is one constant, one use in `canonical_signing_payload()`, and verification recomputes that payload to compare against the stored signature. So every v1 signature stops verifying.

**That break is fine and intended** — redaction has no users yet, and a clean cut beats carrying a compatibility path for zero real signatures. Worth stating in the PR body so the decision is recorded rather than inferred.

What does need fixing is the comment three lines above the change, which still reads:

> Bumping this versions the payload format itself; old signatures with the old prefix continue to verify exactly as they did when written.

That is now untrue, and it is exactly the kind of comment someone will rely on the *next* time this tag moves — at which point there may well be signatures in the field. Either delete the guarantee or replace it with what actually happens: bumping the tag invalidates every signature written under the previous tag, and any future bump needs a verify-side dispatch on the stored prefix if compatibility is wanted.

Worth noting for that future change: the failure presentation is harsh. A stale-format redaction doesn't surface as "old format, re-sign" — it fails through `verify_wire_redaction` as a `WireRejection`, and `verified_redactions_bytes_for_wire` refuses the push. To an operator that reads as metadata tampering, not a version migration. Fine with no data in the field; not fine later.

## Confirm: does tapestry actually distribute client identities?

Purge now re-signs with the active *client* identity, so a receiver must trust the purger's client key — a second enrollment beyond the signing key. The integration test does it by hand and the comment says:

> In production Tapestry distributes trusted client identities; this local-sync fixture enrolls A's public key explicitly.

If that distribution isn't implemented yet, purge propagation passes in the fixture and fails silently in production, with the fixture's manual `redact trust add` masking the gap. Worth confirming before this leaves draft, and linking the issue if it's still pending.

## Process

**25 files across 8 areas in one commit**, titled *"Expand hosted review and agent workflow capabilities"* — which doesn't match the PR title. For security work that costs real things: `git bisect` can't separate the redaction payload change from the clone-planning change, and a later audit can't tell what shipped when.

The release-pipeline hardening in particular is self-contained and unambiguously good. It could land on its own today rather than waiting on the redaction and hosted-sync review.

**`cargo-audit` and `Coverage` are both SKIPPED and `Quality` hasn't concluded.** Probably path filters, but worth a glance on a PR that is explicitly about security — a skipped audit on a security change is the one skip that shouldn't be assumed benign.

thethorne48:
Following up on my open question about trusted client identities — I checked, and the answer is more interesting than a yes/no.

Tapestry does distribute self-sovereign client identities. But nothing carries them into the list this PR depends on. `trusted_keys` appears in six files across heddle and **every writer is `heddle redact trust add` or a test** — no code path populates it from any distribution. The command's own docstring describes the model plainly:

> Add an operator public key to `[redact] trusted_keys` in `.heddle/config.toml`. Subsequent `heddle pull` or `heddle clone` invocations will accept signed redactions from that key.

There are in fact three unrelated trust stores in play, and the one governing purge is the hand-edited one:

| Store | Populated by |
|---|---|
| `[redact] trusted_keys` (`repo_config.rs:68`) | `heddle redact trust add`, by hand |
| `DescriptorKeyring` (`hosted/bootstrap.rs:33`) | hosted distribution — but for Iroh endpoint descriptors, "keyed independently from Iroh endpoint and hosted capability identities" |
| weft biscuit capabilities | weft, at mint time |

One clarification that matters for which key would be distributed: `passkey-support.ts` is a capability detector — it checks `isSecureContext` and `PublicKeyCredential` availability. The key that signs is the Ed25519 WebCrypto device key (`device-key.ts:181`). So the passkey anchors the human and the device key is the signable material; only the latter could land in a receiver's trust list.

So the concern stands as written: purge propagation requires every receiving repo to have manually run `redact trust add` for the purger's client key, which is exactly what the fixture does by hand.

**Not a blocker for this PR** — there are no customers and redaction is unused. But rather than wiring identity distribution into the TOML list, the better fix is to stop having a second trust mechanism at all: weft already models `right("repo", $p, "purge") trusting authority` with no inheritance and no delegation, and has tests proving a forged delegated right doesn't grant it. Tracked in HeddleCo/weft#836, including the constraint that it must verify offline so local peer-to-peer sync — the topology this PR's own integration test uses — keeps working.

Worth linking #836 from this PR so the manual-trust step reads as known-temporary rather than as the intended design.

forge-zephyr:
Corrected the repository-signing doc comment: weft is the authorization source for destructive and state-visibility operations, while the client retains the signing key, so weft cannot forge those operations. The comment now also identifies the local trusted-key list as temporary and references HeddleCo/weft#836 for the superseding authorization model.

forge-zephyr:
Resolved against current `main` and pushed normally (no force-push).

- Merge commit: `4d0d4d5ccaae2d633a19a23ee60ac8d16b8cd090`
- Parents: prior #1129 head `988e502c791de2daf9d75222c908c0c5bbd0ec21` + `main` `d9fc4f1f5873e0663f329b188fc801cc6620bcff`
- Verified immediately before push that the published #1129 head had not moved; push fast-forwarded `988e502c..4d0d4d5c`.

Resolution: I dropped #1129's explicit-spec-first reordering. A CLI spelling such as `127.0.0.1:8421` cannot simultaneously mean “always explicit” and “the configured remote with this exact name” without new syntax; the established named-remote-first behavior is required to retain configured policy. The production resolver therefore keeps `main`'s named-remote precedence and repository-aware `parse_target_for_repository`, including native HTTPS handling. I also removed the dropped alias-shadowing claim from the changelog.

The exact regression `configured_host_port_remote_honors_insecure_policy` was added and observed failing first against #1129's ordering: it resolved the ad-hoc port `8421` instead of the configured target port `9999`. It passes after the resolution, including `insecure == true` and the configured credential key/target.

Verification:

- `cargo test --locked -p heddle-cli-shared remote::tests -- --nocapture`: 3 passed, including the new host:port/insecure regression and native HTTPS round-trip.
- `cargo test --locked -p heddle-cli-shared native_ -- --nocapture`: 3 passed (native parser/source-authority coverage).
- `https_transport_follows_repository_source_authority`: passed.
- `missing_iroh_endpoint_is_named_and_never_downgrades`: passed; the missing Iroh endpoint remains a named hard refusal.
- #1135 compatibility: an isolated, no-commit merge applied cleanly and `cargo test --locked -p heddle-cli --test git_overlay_tls -- --nocapture --test-threads=1` passed all 3 tests. #1135's in-review files were not added to this branch.
- Local exact CI build passed: `cargo build --locked --workspace --tests --bin heddle --bin heddle-fuse-worker`.
- Local exact CI Clippy passed: `cargo clippy --locked --workspace --lib --bins --tests --examples -- -D warnings`.
- The local exact workspace test command reproduced the managed sandbox's documented read-only `.git` fixture failures and stalled in `stdout_stderr_split::json_mode_keeps_stdout_clean_on_every_catalog_command`; it was not reported as a pass. The focused semantic tests above passed.
- Authoritative GitHub CI on `4d0d4d5c` is green: Quality (exact affected-package build, Clippy, and tests), Feature contracts, release pipeline contracts, platform smokes, and final `Check & test` all passed. This clean-filesystem run covers #1129's 413 core tests and signed-purge propagation regression.

Explicitly: a configured remote's `insecure` flag still applies when the remote name itself parses as a bare `host:port`.

thethorne48:
Landing this. The conflict that had it stuck is resolved in `4d0d4d5c` via HeddleCo/heddle#1137, and the resolution took the better of the two options: **#1129's explicit-spec-first reordering was dropped as incidental scope**, keeping main's named-remote precedence and #1133's native HTTPS discovery.

That mattered because the reordering deleted the branch whose comment reads *"Named remote first so configured `insecure` applies even when the name also happens to parse as a bare host:port"*, and the replacement path returned `insecure: false` unconditionally. In a PR titled *harden security controls*, that would have failed open — quietly. There is now a regression test for exactly that case, observed failing first (`8421` resolving instead of the configured `9999`).

Also credited: the session ran an isolated compatibility merge against HeddleCo/heddle#1135 and confirmed all three `git_overlay_tls` tests pass together. Three heddle PRs are in flight at once, so checking that was worth doing and I had not asked for it.

**One forward-looking note.** This touches `repository_redaction.rs`, `repository_signing.rs`, and `repository_state_visibility.rs` — the exact surfaces HeddleCo/weft#836 step 2 will change when redact/purge/visibility move to offline-verifiable owner-anchored capabilities. Nothing here conflicts with that direction, but whoever picks up step 2 should read this PR first rather than assuming those files are untouched since the model doc was written.

Issue #1132, opened 2026-07-28
Git lane ignores every CA trust mechanism, including the one the hosted lane honours


thethorne48:
The Git-Overlay push/pull lane verifies TLS against a trust set that **no operator configuration can change**. A server presenting a certificate from a private CA is unreachable over that lane, and there is no documented or undocumented way to make it reachable.

## Reproduction

Against a weft dev stack serving a certificate signed by a CA not in the public roots:

```text
$ heddle push origin
Error: Push failed for origin: git error: io error: HTTP request to
https://127.0.0.1:8431/demoalice/demo/info/refs?service=git-upload-pack failed:
io: invalid peer certificate: UnknownIssuer
```

All three standard escape hatches were tried, and **none** of them changes the outcome — the error is byte-identical each time:

| Mechanism | Result |
|---|---|
| `HEDDLE_REMOTE_TLS_CA_CERT=<pem>` | `UnknownIssuer` |
| Certificate installed via `update-ca-certificates` (system trust store) | `UnknownIssuer` |
| `SSL_CERT_FILE=<pem>` | `UnknownIssuer` |
| `SSL_CERT_FILE=/etc/ssl/certs/ca-certificates.crt` | `UnknownIssuer` |

`HEDDLE_REMOTE_TLS_CA_CERT` is the sharpest part of this: it **works** for the hosted bootstrap path in the same binary, in the same process invocation. With it set, `heddle auth login` against the same server fetches `/.well-known/heddle/iroh-descriptor-key`, pins descriptor trust, and authenticates successfully. The very next command against the same host over the Git lane fails on the certificate.

So the variable is not ignored globally — it is honoured by one transport and silently not consulted by the other. A user who has just watched it work has no reason to suspect it is not in effect.

## Why this matters beyond dev stacks

Any self-hosted or enterprise deployment behind a private/internal CA is affected. The hosted lane can be configured to trust it; the Git lane cannot be configured at all. That is not a dev-only inconvenience — it silently rules out a deployment topology.

It also blocks driving the full clean-stack flow end to end, which is how it was found.

## What to establish first

Where the Git lane's HTTPS client gets its roots, and whether that is deliberate. If it is pinned to compiled-in roots on purpose, say so in the error and in `heddle help remotes`, because the current message is indistinguishable from a misconfigured server.

## Acceptance

- The Git lane honours the **same** CA configuration the hosted lane already honours. One mechanism, both transports.
[...]

PR #1134, opened 2026-07-29
fix: accept HTTPS for native remotes

### LINKED ISSUE #1133: accept https:// for native remotes; heddle:// carries no addressing information
`heddle://` carries no addressing information. In `crates/cli-shared/src/remote/target.rs`, the parser strips it and discards it:

```rust
fn parse_network_with_repo_path(s: &str) -> Option<(SocketAddr, Option<String>)> {
    if let Some(rest) = s.strip_prefix("heddle://") {
        return parse_network_with_repo_path(rest);
    }
    if let Ok(addr) = s.parse::<SocketAddr>() { ... }
    ...
}
```

Everything after that resolves a plain `host:port` plus a repo path. `Display` re-adds the prefix on the way out. **Iroh does not need a custom scheme** — the scheme is a classification token that tells the CLI "the user meant the native lane," not an addressing scheme.

Meanwhile `https://<same host>/<same path>` names the *same repository on the same port*, and routes to the Git lane instead.

## Why this costs real time

Driving the clean-stack demo, the failure sequence was:

1. `heddle remote add origin https://127.0.0.1:8431/demoalice/demo` — **accepted silently**.
2. `heddle push origin` — fails with `invalid peer certificate: UnknownIssuer` from a `git-upload-pack` URL.

That reads as a server or TLS misconfiguration. It is actually "you selected a different transport." Nothing in the flow says so. I then tried the schemeless form and got *"is a Git remote, not a Heddle-native remote"* — pointing at `heddle clone`, still never naming the scheme that would have worked.

Cost roughly an hour, and a subsequent agent hit the same wall independently on HeddleCo/weft#866 before switching to `heddle://`. Two independent parties, same trap.

## The proposal

**Accept `https://` in a native Heddle repo and resolve the transport by asking the server.**

The client already fetches `https://<host>/.well-known/heddle/iroh-endpoint` during `auth login` — that is where descriptor trust is discovered and pinned. So an `https://` origin is already sufficient to learn that the server speaks Iroh, over the exact channel where trust is established. The scheme is doing work that discovery does better and earlier.

Keep `heddle://` as an explicit override for forcing the native lane or skipping discovery.

Note the dispatch is already documented as repo-mode-driven — `heddle help remotes` says *"Remotes — dispatched by repository source authority."* The scheme requirement is an additional guard layered on top of that, not the mechanism itself.

## What must be decided, not assumed

**What happens when an `https://` server does not advertise an Iroh endpoint.** Silently falling back to the Git lane would be a transport downgrade chosen on the user's behalf, and given HeddleCo/heddle#1132 it would also land them on the lane that cannot do private-CA TLS. It should fail with a message naming what was missing.

Get that wrong and this change trades a confusing error for a silent downgrade, which is worse.

## Acceptance

- `heddle remote add` with `https://` in a native repo resolves to the native lane via well-known discovery, and pushes.
- `heddle://` still works and still forces native.
- A server with no advertised Iroh endpoint produces a **clear refusal**, not a silent Git-lane fallback — shown failing first.
- Where a remote is genuinely a Git remote, the error names the scheme that would have worked instead of only pointing at `heddle clone`.

## Related

- HeddleCo/heddle#1132 — the Git lane's TLS trust gap, which is what made the mis-dispatch surface as a certificate error
- HeddleCo/weft#866 — the acceptance push that independently hit this

forge-zephyr:
## Summary

- Treat `https://` as a native network target only when repository source authority is native; Git-overlay classification stays unchanged, `heddle://` remains the explicit force-native form, and stored HTTPS URLs round-trip unchanged.
- Reuse the hosted signed endpoint-descriptor discovery and trust path before adding a native HTTPS remote. If `/.well-known/heddle/iroh-endpoint` is absent, refuse with structured advice and do not fall back to Git: that fallback would be an implicit transport downgrade.
- Name `heddle://<host>/<repo>` in the genuine-Git-remote mismatch advice. This does not change Git-lane CA handling from #1132.

## Closes

Closes #1133

## Test evidence

- Negative test was observed failing first with the old generic `InvalidDescriptor` 404, then passing as `missing_iroh_endpoint_is_named_and_never_downgrades`.
- Focused parser, source-authority routing, HTTPS round-trip, missing-endpoint, and Git-remote advice tests pass.
- Live local hosted verification: `remote add` preserved the HTTPS spelling and a real push returned `transport: heddle`, `status: pushed`; the explicit `heddle://` form also pushed successfully.
- Live 404 verification exits 74 with `kind: native_https_iroh_endpoint_missing`, names `/.well-known/heddle/iroh-endpoint`, and states that no Git fallback is attempted.
- CI affected-package `cargo build --locked ... --tests --bin heddle --bin heddle-fuse-worker` and `cargo clippy --locked ... --lib --bins --tests --examples -- -D warnings` pass.
- All four CI alternate-feature `cargo check --locked ... --no-default-features --features ...` commands pass (the existing alternate-feature dead-code warnings remain).
- The exact affected-package `cargo test --locked ...` command was also run. In this managed sandbox its integration fixtures see injected read-only `.git` mounts, producing broad unrelated failures; the run later stopped making progress in `stdout_stderr_split::json_mode_keeps_stdout_clean_on_every_catalog_command` and was interrupted after the failures were recorded. GitHub CI remains the authoritative full-suite run.

---
<a href="https://app.blacksmith.sh/HeddleCo/codesmith/heddle/pr/1134"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1787885986&installation_model_id=21489&pr_number=1134&repository=HeddleCo%2Fheddle&return_to=https%3A%2F%2Fgithub.com%2FHeddleCo%2Fheddle%2Fpull%2F1134&signature=d61606b00dd387322933815aa398ca7fd2b58a4a1f5af3f0a3b1c2bac1cd4540"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/crates/cli-contract/src/cli/commands/advice.rs
+++ b/crates/cli-contract/src/cli/commands/advice.rs
@@ -1026,7 +1026,7 @@ impl RecoveryAdvice {
             format!(
                 "Refusing to {action}: remote '{remote}' is a Git remote, not a Heddle-native remote"
             ),
-            "Use a Heddle-native remote here, or clone/adopt that Git remote in a Git-overlay checkout.",
+            "For a native server, use `heddle://<host>/<repo>` to force Heddle transport, or clone/adopt this Git remote in a Git-overlay checkout.",
             format!("remote '{remote}' resolves to Git storage"),
             format!(
                 "{action} would route a Git repository through Heddle-native sync and fail after setup work"
@@ -1037,6 +1037,23 @@ impl RecoveryAdvice {
                 "heddle clone <remote> <fresh-path>".to_string(),
                 "heddle remote add <name> <url>".to_string(),
             ],
+        )
+    }
+
+    pub fn native_https_iroh_endpoint_missing(action: &str, remote: &str) -> Self {
+        Self::safety_refusal(
+            "native_https_iroh_endpoint_missing",
+            format!(
+                "Refusing to {action}: HTTPS server for remote '{remote}' does not advertise an Iroh endpoint"
+            ),
+            "Configure the server to publish `/.well-known/heddle/iroh-endpoint`, then retry. Heddle will not silently fall back to Git transport.",
+            format!(
+                "remote '{remote}' returned no Iroh endpoint descriptor at `/.well-known/heddle/iroh-endpoint`"
+            ),
+            "falling back would silently downgrade a native repository to Git transport",
+            "remote configuration, Heddle refs, Git refs, and worktree files were left unchanged",
+            "heddle help remotes",
+            vec!["heddle help remotes".to_string()],
         )
     }
 
--- a/crates/cli-shared/src/remote/mod.rs
+++ b/crates/cli-shared/src/remote/mod.rs
@@ -12,7 +12,7 @@ use std::{
 };
 
 use objects::fs_atomic::write_file_atomic;
-use repo::Repository;
+use repo::{Repository, RepositoryCapability};
 use serde::{Deserialize, Serialize};
 pub use target::RemoteTarget;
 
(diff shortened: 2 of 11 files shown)


thethorne48 (approved):
Approving. The decision I flagged as the one that decides whether this is an improvement went the right way: a server that does not advertise an Iroh endpoint gets a **hard refusal**, not a Git-lane fallback. The error names what was missing and states `preserved` and `would_change` explicitly, so the user learns the transport was the problem rather than guessing at TLS.

Silent fallback would have traded a confusing error for a silent downgrade — and given HeddleCo/heddle#1132 it would also have dropped the user onto the lane that cannot do private-CA TLS. Refusing is worth more than convenience here.

Three other things done right:

- **The negative test was shown failing first** (`missing_iroh_endpoint_is_named_and_never_downgrades`), so it is keyed on the behaviour rather than passing by construction.
- **The entered URL spelling is preserved** in `.heddle/remotes.toml` rather than silently rewritten to `heddle://`. A remote that comes back different from what was typed is its own small betrayal of trust.
- The Git-remote mismatch error now **names the working native scheme** instead of only pointing at `heddle clone`. That is the specific gap that cost me an hour driving the clean-stack demo, and cost the weft#866 agent the same before it switched.

The parser finding is confirmed as reported: `heddle://` was only ever a classification token, discarded before addressing.

PR #1135, opened 2026-07-29
fix(remotes): honor configured CA in the Git HTTPS lane

### LINKED ISSUE #1132: Git lane ignores every CA trust mechanism, including the one the hosted lane honours

thethorne48:
## Summary

- The Git-Overlay push/pull lane now honours the **same** CA configuration the hosted lane already honours, so one setting works across both transports.
- Root cause was in the dependency, not here: Sley's rustls-backed `UreqHttpClient` verified against compiled-in Mozilla `webpki-roots` and consulted nothing else. It was a default, not a deliberate Heddle policy. Fixed in HeddleCo/sley#189.
- Adds a private-CA integration test and a support harness that stands up a real HTTPS Git server.
- Failure messages now name the configuration that would fix them rather than only surfacing the rustls error class.

## Why this was worth chasing

The defect made any self-hosted or enterprise deployment behind a private CA unreachable over the Git lane, with **no configuration that could change it**. All four standard escape hatches produced byte-identical `UnknownIssuer`: `HEDDLE_REMOTE_TLS_CA_CERT`, the system trust store via `update-ca-certificates`, `SSL_CERT_FILE=<pem>`, and `SSL_CERT_FILE=<system bundle>`.

The sharpest part was that `HEDDLE_REMOTE_TLS_CA_CERT` **worked for the hosted bootstrap path in the same binary and the same process invocation** — `auth login` would pin descriptor trust and authenticate, and the very next command over the Git lane would fail on the certificate. Honoured by one transport, silently never consulted by the other, so a user who had just watched it work had no reason to suspect it was inert.

## Closes

Closes HeddleCo/heddle#1132

## Test evidence

Deliberately failing run before the fix, with `HEDDLE_REMOTE_TLS_CA_CERT` set:

```text
cargo test --locked -p heddle-cli --test git_overlay_tls \
  git_overlay_pull_honours_remote_tls_ca_cert -- --nocapture --test-threads=1

exit 101: invalid peer certificate: UnknownIssuer
```

After the fix:

```text
private-CA pull, push, and actionable-failure tests: 3 passed
shared client-config tests:                          10 passed
git-projection unit tests:                           57 passed
```

## Dependency ordering

**HeddleCo/sley#189 must merge first.** `Cargo.lock` currently pins that branch's commit `eb4798a6`; once it lands the pin moves to the merged SHA. Merging this first would pin a branch that no longer exists.

---
<a href="https://app.blacksmith.sh/HeddleCo/codesmith/heddle/pr/1135"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1787893024&installation_model_id=21489&pr_number=1135&repository=HeddleCo%2Fheddle&return_to=https%3A%2F%2Fgithub.com%2FHeddleCo%2Fheddle%2Fpull%2F1135&signature=eb9e6d3fa07a4269551e046af45094a41dc7bef2a77acdc9513d86bc48b92b5c"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


thethorne48:
## cargo-deny is failing for a structural reason, not a flaky one

```text
error[source-not-allowed]: detected 'git' source not explicitly allowed
```

repeated across every sley crate. The cause is a real policy conflict, and it changes what has to happen next.

**main depends on published crates:**

```toml
sley = { version = "0.5.0", features = ["mmap", "fast-sha1"] }
sley-transport = "0.5.0"
```

**this branch pins a git revision:**

```toml
sley = { git = "https://github.com/HeddleCo/sley", rev = "eb4798a6...", features = [...] }
sley-transport = { git = "https://github.com/HeddleCo/sley", rev = "eb4798a6..." }
```

heddle's supply-chain policy is published-crates-only, and cargo-deny enforces it. So this is not a pin that needs repointing at the merged SHA — **a git dependency cannot merge here at all**, whichever commit it names. That is a deliberate control and the right one; the PR simply cannot satisfy it in its current shape.

### The actual chain

1. HeddleCo/sley#189 merged to sley . ✅
2. **sley needs a release** carrying the CA fix. Workspace is currently at ; this wants a  (additive, non-breaking — it adds a configurable CA source and keeps the bundled Mozilla roots as the default set).
3. This PR then changes to  and drops the git pin entirely.

Step 2 is the blocker and it is an operator action — sley has , and heddle's own release/publish pipelines were just hardened in HeddleCo/heddle#1129, so the machinery is fresh.

### Meanwhile

The heddle-side work itself is sound and CI-green apart from this: the failing-first test, the private-CA integration harness, and the actionable error messages all stand. Nothing needs rewriting — it needs a dependency that exists on crates.io.

Marking this blocked on the sley release rather than churning the branch.

thethorne48:
Correcting the previous comment — inline code spans were eaten by my shell before posting. The tail of it should read:

### The actual chain

1. HeddleCo/sley#189 merged to sley `main`. ✅
2. **sley needs a release** carrying the CA fix. The workspace is currently at `0.5.0`; this wants a `0.5.1` — additive and non-breaking, since it adds a configurable CA source and keeps the bundled Mozilla roots as the default trust set.
3. This PR then depends on `sley = "0.5.1"` and `sley-transport = "0.5.1"`, dropping the git pin entirely.

Step 2 is the blocker and it is an operator action. sley has `.github/workflows/publish-crates.yml`, and heddle's own release and publish pipelines were just hardened in HeddleCo/heddle#1129, so the machinery is current.

thethorne48:
**Blocked on a regression in the release this was waiting for.** No changes pushed.

sley 0.5.1 published today, and the upgrade attempt immediately found:

> sley's worktree/ignore traversal **escapes temporary repository roots** and scans unrelated `/home/scratch` directories. It eventually encounters another sandbox-owned `.heddle` directory and fails with EACCES. `strace` confirmed the out-of-repository traversal.

Filed as HeddleCo/sley#202 and dispatched. The EACCES is incidental — **the traversal escaping the repository root is the defect**, and it would be silent on any machine where the parent directories happen to be readable, which is most of them.

## What did verify on the dependency swap

Worth recording, because it means only the regression is in the way:

- `cargo deny check all` **passed** — advisories, bans, licenses, sources. That was the whole reason this PR could not merge, and the published dependency resolves it.
- Private-CA integration: 3/3 passed against the published crate.
- CI-shaped workspace clippy passed with `-D warnings`.
- Grep confirms no `sley` git source remains.

## One API change to carry forward

0.5.1 adds a required `FetchOptions::negotiate_only` field. Setting it to `false` at the four existing fetch call sites preserves prior behaviour — that work is done and sitting uncommitted alongside the dependency change.

## Next

0.5.2 with the traversal fix, then this lands. The brief for this task said to treat any behaviour change beyond the CA fix as a finding rather than something to adjust around — which is exactly what happened, and why this is a blocked PR rather than a merged one with a mysterious workaround in it.

PR #1138, opened 2026-07-29
feat(client): add inert owner-anchored authorization surface

### LINKED ISSUE #1136: client surface for owner-anchored authorization (weft#836 step 0b)
Step 0b of HeddleCo/weft#836's cutover. The contract landed in HeddleCo/api#57; this is the **heddle client surface** for it.

**Read [the contract design](https://github.com/HeddleCo/api/issues/55#issuecomment-5112271818) and HeddleCo/api#56 before starting.** The shape is settled and reviewed; this issue implements against it, it does not redesign it.

## Scope

The client side of the six contract objects, in `heddle`:

1. **Owner root** — generate, store, and load an offline-verifiable owner key.
2. **Bootstrap** — WebAuthn-approved human bootstrap, and agent-key bootstrap for deferred accounts.
3. **Rotation and recovery** — construct and sign `OwnerKeyTransition` for `ROTATE`, `RECOVER`, `RECOVERY_POLICY`, and `CLAIM_DEFERRED_HUMAN`, including the guardian threshold.
4. **Owner-signed capability submission** — the client-originated replacement for asking weft to mint.
5. **Anonymous authentication** — the self-authenticated client key form.
6. **Clone keyring persistence** — store the signed root, transition history, pinned state, and public policy so a clone can verify **offline**.

## The constraint that governs everything

**Inert until weft cuts over.** These surfaces must not let a client obtain owner-anchored authorization while weft still accepts server-minted credentials. Two live authorization paths means the weaker one decides, which is the precise hazard weft#836 is structured around.

api#57 achieved inertness *structurally* — the proto declares no service or RPC and nothing imports it. Aim for the same quality of guarantee here rather than a documented intention: **it should be difficult to wire this up early by accident, not merely discouraged.** State in the PR how each surface stays inert and what would make it live.

## Where the offline proof lives

weft can define the clone-keyring bytes but **cannot prove a real filesystem clone pins and uses them offline** — that proof belongs here. This is the repo where "verification works with no network" becomes demonstrable rather than asserted, and weft#836's audit called it out specifically.

So: a test that clones, disconnects, and reaches both a correct **allow** and a correct **deny** with no network at all. A design that only works while weft is reachable will pass a naive suite, because the test environment always has weft reachable.

## Recovery specifics already decided (HeddleCo/weft#873)

- Threshold ≥2 by default, with a **device-independent** co-factor — paper kit or social guardian. Device-resident co-factors do not count; recovery exists for when devices are gone.
- Weft-only (1-of-1) is permitted but must be **behind an explicit confirmation**, and the client should refuse to construct it silently. That configuration makes the account custodial.
- Guardian provenance is on the wire as `RECOVERY_GUARDIAN_KIND_{PAPER,SOCIAL,WEFT}` — preserve it. A weft guardian being indistinguishable from a paper one defeats the threshold rule.
- The v1 paper recovery kit is a client-generated 256-bit Ed25519 seed encoded for paper/QR. Only its public key goes on the wire.

## Out of scope

- The tapestry client surface (separate issue).
- weft's cutover (weft#836).
- Offline revocation (weft#872) — ships open, with short capability TTLs. Do not solve it here and do not design as if solved.

## Verification

- Round-trip every contract object through construction, signing, serialization, and verification.
- **Offline allow and offline deny from a real clone, network genuinely down.**
- Rotation: rotated key verifies, previous key does not after the handover window.
- Recovery: threshold satisfied verifies; one signature short is refused.
- **Negative cases, each shown failing before it passes:**
  - a 1-of-1 weft policy is not constructible without the explicit confirmation path;
  - a capability asserting a right outside its granted set does not verify;
  - a transition with a broken `previous_state_hash` chain is refused;
  - guardian kind cannot be forged to make a WEFT guardian look like PAPER.

## Related

- HeddleCo/api#55, #56 — the contract design and implementation
- HeddleCo/weft#836 — the cutover this unblocks
- HeddleCo/weft#873 — recovery policy decisions
- HeddleCo/weft#863 — deferred accounts, downstream of all of it

forge-zephyr:
## Summary

Implements the client-side api#57 owner-authorization contract as data/crypto primitives: the owner root, human and deferred bootstrap objects, all four rotation/recovery transitions, owner-signed capability chains plus subject Biscuits, anonymous self-authentication, and clone keyring persistence/offline evaluation. The wire containers mirror the reviewed protobuf fields and tags, while signatures use fixed-order, length-prefixed canonical encodings rather than protobuf bytes.

Recovery construction defaults to threshold 2, requires a paper or social co-factor whenever Weft participates, preserves PAPER/SOCIAL/WEFT provenance inside the signed policy, and exposes 1-of-1 Weft custody only through the explicit confirmation type. Paper kit v1 is an exact client-generated 32-byte Ed25519 seed; only its public key enters the wire policy. Numeric TTL ceilings remain explicit `VerificationLimits` inputs because api#55 deliberately leaves their values as cutover product inputs.

Clone keyrings are bounded, canonically re-encoded, chain-verified, pinned once, and atomically persisted at `.heddle/owner-authorization/keyring.pb`. The offline authorizer accepts no network handle and binds evaluation to the pinned clone path as well as the signed spool selector.

No Cargo manifest, `Cargo.lock`, API dependency pin, TLS, or CA configuration changed, keeping this patch outside #1135.

### Inertness by surface

- Owner root: only the new data/crypto module can construct or inspect it. No ownership lookup, trust list, hosted client, CLI, core, or repository verifier imports it. It becomes live only when the exclusive post-cutover verifier consumes `owner_id` after the server-minted path is removed.
- Bootstrap: request/receipt types have no RPC or hosted transport path; the response is checked only as an unsigned acknowledgement and cannot yield a bearer. It becomes live when the bootstrap route is added in the atomic Weft cutover.
- Rotation and recovery: transitions are pure signed objects applied only to an explicitly supplied `VerifiedOwnerState`; they cannot mutate `device_roots` or current credential authority. They become live when the exclusive owner-state verifier applies them.
- Capability submission: the bundle and submission receipt are data-only. Nothing copies them into `grant_envelope`, `CallContext`, or a bearer, and no current handler reads them. They become live when an owner-only submission/verifier replaces minting.
- Anonymous auth: the self-signed credential and continuity request have no endpoint or current verifier integration. They become live only when the exclusive anonymous verifier replaces server-minted anonymous credentials.
- Clone keyring: persistence uses a new path that current clone, trusted-key, redact, and purge receivers do not read. It becomes live only when the later clone/receiver cutover imports this module and switches to it without a legacy fallback.

`owner_authorization_inertness.rs` enforces the structural boundary by scanning production client, CLI, core, and repository sources for any current import and by rejecting known live credential edges inside the new module.

## Closes

Closes #1136

## Test evidence

- `rustfmt +nightly --edition 2024 <touched files>` on touched Rust files only.
- `cargo test --locked -p heddle-client --lib --test owner_authorization_contract --test owner_authorization_inertness`: 137 library tests, 13 contract tests, and the inertness tripwire pass on rebased current main.
- `cargo test --locked -p heddle-cli --test owner_authorization_offline_clone -- --nocapture`: passes after a real `heddle init`, `capture`, and filesystem `clone`; the test then installs a per-thread seccomp filter denying socket/socketpair/connect/bind/listen/accept/send/recv/shutdown syscalls before loading or evaluating the keyring.

```text
CLONE_KEYRING=/home/scratch/.tmpcqetnA/clone/.heddle/owner-authorization/keyring.pb
NETWORK_PROBE=blocked errno=EPERM syscall=socket
OFFLINE_ALLOW=READ path=acme/heddle/src
OFFLINE_DENY=WRITE path=acme/heddle/src
OFFLINE_DENY=READ path=acme/sibling reason=pinned-clone
```

- Exact selected-package build passes: `cargo build --locked -p heddle-cli -p heddle-cli-args -p heddle-client -p heddle-cli-contract -p heddle-cli-render --tests --bin heddle --bin heddle-fuse-worker`.
- Exact selected-package lint passes: `cargo clippy --locked -p heddle-cli -p heddle-cli-args -p heddle-client -p heddle-cli-contract -p heddle-cli-render --lib --bins --tests --examples -- -D warnings`.
- CI feature checks pass for `heddle-repo` with `git-overlay,zstd` and `native,zstd`, and for `heddle-cli` with `git-overlay,client,semantic,zstd` and `native,semantic,zstd`, all with `--no-default-features --locked`.
- `bash scripts/check-facade-render-free.sh` passes; 12 script unit tests pass.
- Required negative assertions reject implicit 1-of-1 Weft custody, capability widening, a broken `previous_state_hash`, and WEFT-to-PAPER guardian-kind rewriting. Threshold recovery succeeds at 2 signatures and rejects one signature short; rotation accepts the new key and expires the old key after the handover.

The broad affected-package cargo test command was also watched locally. Existing native CLI temp-directory cases fail in this managed sandbox because read-only injected ancestor `/home/scratch/.git` and `/tmp/.git` directories are discovered before those tests run; the new focused client and real-clone tests isolate that environment and pass. PR CI is the authoritative unsandboxed broad-suite result.

---
<a href="https://app.blacksmith.sh/HeddleCo/codesmith/heddle/pr/1138"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1787897175&installation_model_id=21489&pr_number=1138&repository=HeddleCo%2Fheddle&return_to=https%3A%2F%2Fgithub.com%2FHeddleCo%2Fheddle%2Fpull%2F1138&signature=da3bbc83c2f5ddc9ae88de7ab69168f710d913a75e9cf6b37e76ef630ddc80e0"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/crates/cli/tests/owner_authorization_offline_clone.rs
+++ b/crates/cli/tests/owner_authorization_offline_clone.rs
@@ -1,0 +1,268 @@
+// SPDX-License-Identifier: Apache-2.0
+
+#![cfg(target_os = "linux")]
+
+use std::{
+    net::{Ipv4Addr, SocketAddrV4, TcpStream},
+    path::{Path, PathBuf},
+    process::Command,
+    time::Duration,
+};
+
+use heddle_client::owner_authorization::{
+    AuthorizationError, AuthorizationKey, CloneKeyringStore, GuardianSigner, OfflineAuthorizer,
+    OfflineRequest, RecoverySetup, VerificationLimits, create_clone_keyring,
+    create_direct_capability, create_human_owner_root, mint_subject_biscuit, verify_owner_root,
+    wire::{
+        CapabilityPrincipal, CapabilityPrincipalKind, CloneOwnerPinKind, SpoolCapabilityAction,
+        SpoolCapabilityGrant, SpoolSelector,
+    },
+};
+use sley::Repository as SleyRepository;
+use tempfile::TempDir;
+
+const NOW: i64 = 1_000;
+
+fn limits() -> VerificationLimits {
+    VerificationLimits::new(300, 3_600, 1024 * 1024).expect("limits")
+}
+
+fn heddle(args: &[&str], cwd: Option<&std::path::Path>, test_home: &std::path::Path) {
+    let mut command = Command::new(env!("CARGO_BIN_EXE_heddle"));
+    command
+        .env_clear()
+        .args(args)
+        .env("HOME", test_home)
+        .env("HEDDLE_CONFIG", test_home.join("config.toml"))
+        .env("RUST_BACKTRACE", "1")
+        .env("HEDDLE_PRINCIPAL_NAME", "Offline Clone Test")
+        .env("HEDDLE_PRINCIPAL_EMAIL", "offline-clone@heddle.test");
+    if let Some(cwd) = cwd {
+        command.current_dir(cwd);
+    }
+    let output = command.output().expect("run heddle");
+    assert!(
+        output.status.success(),
+        "heddle {args:?} failed\nstdout: {}\nstderr: {}",
+        String::from_utf8_lossy(&output.stdout),
+        String::from_utf8_lossy(&output.stderr)
+    );
+}
+
+fn disconnect_network_for_current_thread() {
+    const LOAD_SYSCALL_NUMBER: u16 = 0x20;
+    const JUMP_EQUAL: u16 = 0x15;
+    const RETURN: u16 = 0x06;
+    const SECCOMP_RETURN_ERRNO: u32 = 0x0005_0000;
+    const SECCOMP_RETURN_ALLOW: u32 = 0x7fff_0000;
+
+    let denied = [
+        libc::SYS_socket,
+        libc::SYS_socketpair,
+        libc::SYS_connect,
+        libc::SYS_bind,
+        libc::SYS_listen,
+        libc::SYS_accept,
+        libc::SYS_accept4,
+        libc::SYS_sendto,
+        libc::SYS_sendmsg,
+        libc::SYS_recvfrom,
+        libc::SYS_recvmsg,
+        libc::SYS_shutdown,
+    ];
+    let mut filters = Vec::with_capacity(denied.len() * 2 + 2);
+    filters.push(libc::sock_filter {
+        code: LOAD_SYSCALL_NUMBER,
+        jt: 0,
+        jf: 0,
+        k: 0,
+    });
+    for syscall in denied {
+        filters.push(libc::sock_filter {
+            code: JUMP_EQUAL,
+            jt: 0,
+            jf: 1,
+            k: syscall as u32,
+        });
+        filters.push(libc::sock_filter {
+            code: RETURN,
+            jt: 0,
+            jf: 0,
+            k: SECCOMP_RETURN_ERRNO | libc::EPERM as u32,
+        });
+    }
+    filters.push(libc::sock_filter {
+        code: RETURN,
+        jt: 0,
+        jf: 0,
+        k: SECCOMP_RETURN_ALLOW,
+    });
+    let program = libc::sock_fprog {
+        len: filters.len().try_into().expect("seccomp filter length"),
+        filter: filters.as_mut_ptr(),
+    };
+
+    // SAFETY: `program` points to `filters` for the duration of both prctl
+    // calls. NO_NEW_PRIVS permits this thread to add a strictly more
+    // restrictive seccomp filter; the filter returns EPERM for networking
+    // syscalls and ALLOW for every other syscall.
+    unsafe {
+        assert_eq!(
+            libc::prctl(libc::PR_SET_NO_NEW_PRIVS, 1, 0, 0, 0),
+            0,
+            "enable no-new-privileges"
+        );
+        assert_eq!(
+            libc::prctl(
+                libc::PR_SET_SECCOMP,
+                libc::SECCOMP_MODE_FILTER,
+                &program as *const libc::sock_fprog,
+            ),
+            0,
+            "install network-denying seccomp filter"
+        );
+    }
+}
+
+fn create_real_clone(temp: &TempDir) -> PathBuf {
+    let source = temp.path().join("source");
+    let clone = temp.path().join("clone");
+    std::fs::create_dir(&source).expect("source directory");
+    SleyRepository::init(&source).expect("isolate source from ancestor Git repositories");
+    heddle(&["init"], Some(&source), temp.path());
+    std::fs::write(source.join("README.md"), "offline clone\n").expect("seed file");
+    heddle(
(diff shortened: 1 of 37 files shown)


thethorne48 (approved):
Approving. This is step 0b of the HeddleCo/weft#836 chain and it delivered the two things that were hard to get right.

## Inertness is structural, and it is tested

**+4,414 lines, 0 deletions across 37 files.** The only change to an existing file is two lines in `crates/client/src/lib.rs` declaring the module. Nothing existing was modified to accommodate it, which is inertness close to by construction — there is no route to reach these surfaces because nothing calls them.

Better still, `owner_authorization_inertness.rs` scans production client, CLI, core, and repository code for accidental imports, routes, hosted-client implementations, and existing authorization integration points. That converts inertness from a property someone has to remember into one **CI notices when it breaks** — which is what I asked for and the thing that would otherwise decay silently as the cutover approaches.

The per-surface accidental-live analysis is precise about the trigger in each case ("becomes live when submission verification replaces server minting", "when clone/import switches to it without legacy fallback"). That is the checklist whoever performs the atomic cutover will need.

## The offline proof is real

This is the part only this repo could establish, and it was the audit's specific ask on weft#836.

```text
NETWORK_PROBE=blocked errno=EPERM syscall=socket
OFFLINE_ALLOW=READ  path=acme/heddle/src
OFFLINE_DENY=WRITE  path=acme/heddle/src
OFFLINE_DENY=READ   path=acme/sibling reason=pinned-clone
```

**Socket creation blocked by seccomp**, not merely unused — so this demonstrates verification working with no network rather than a code path that happens not to dial. That distinction is the whole reason the test was worth demanding: a design that only works while weft is reachable passes a naive suite, because the test environment always has weft reachable.

And it reaches a correct **allow** and two distinct correct **denies**, including one that fails for the pinned-clone reason rather than generically.

## Also correct

Stayed out of `Cargo.lock` and TLS/CA configuration as instructed, so no collision with HeddleCo/heddle#1135. Touched-file-only nightly rustfmt. All four required negative cases pass, including guardian-kind forgery and the broken `previous_state_hash` chain.

## What this unblocks and what it does not

Merging this leaves the chain at: **api contract ✅ → heddle client ✅ → tapestry client → weft#836 atomic cutover → weft#863.**

Nothing here is reachable until that cutover, by design. Worth restating because a large additive PR landing on main can read as "owner-anchored auth is live" — it is not, and the inertness test is what keeps that true.

PR #1140, opened 2026-07-29
fix(auth): forbid attenuated purge capabilities

forge-zephyr:
## Summary

- reject `purge` in every attenuated owner-capability child
- retain exact-spool `purge` in a direct owner-signed capability
- add a regression that builds a validly signed delegated-purge chain and requires rejection

## Closes

Supports HeddleCo/weft#248.

## Test evidence

- `rustfmt +nightly --edition 2024 crates/client/src/owner_authorization/capability/scope.rs crates/client/src/owner_authorization/capability/tests.rs`
- `cargo test --locked -p heddle-client --lib --test owner_authorization_contract --test owner_authorization_inertness` — 138 library, 13 contract, and 1 inertness test passed
- `cargo clippy --locked -p heddle-client --lib --tests -- -D warnings` — passed

The pre-fix owner-chain conformance seed in HeddleCo/weft#248 accepts this exact delegated-purge chain; the fixed verifier rejects it while its direct owner parent still verifies.

### DIFF
--- a/crates/client/src/owner_authorization/capability/scope.rs
+++ b/crates/client/src/owner_authorization/capability/scope.rs
@@ -134,6 +134,13 @@ pub(crate) fn grant_covers(
     parent: &[SpoolCapabilityGrant],
     child: &[SpoolCapabilityGrant],
 ) -> bool {
+    if child.iter().any(|grant| {
+        grant
+            .actions
+            .contains(&(SpoolCapabilityAction::Purge as i32))
+    }) {
+        return false;
+    }
     child.iter().all(|child_grant| {
         let Some(child_selector) = child_grant.spool.as_ref() else {
             return false;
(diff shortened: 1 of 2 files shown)


thethorne48 (approved):
Approving. This closes a real vulnerability that HeddleCo/weft#248's cross-client fuzzer found: **an attenuated child `purge` was accepted under an exact `grant + purge` parent.**

Purge is not delegable by attenuation — it must come from the authority block, because purge destroys bytes irreversibly. That invariant has been stated since the rule pack was written, and this client did not enforce it.

Two things make the fix trustworthy rather than blunt:

- The regression builds a **validly signed** delegated-purge chain and requires rejection. Not a malformed input — a well-formed one that should be refused on authority grounds.
- **Direct owner-signed exact-spool purge still verifies.** A fix that rejected delegated purge by rejecting purge generally would pass the same test and break the feature.

Worth recording how this was found: tapestry had the identical defect, independently. Two separate implementations of one contract drifted the same way, which no single-implementation test would have caught. That is the entire argument for cross-client fuzzing, and it paid for itself on the first run.

Issue #1144, opened 2026-07-30
[p0] data loss: heddle context set orphans existing git history in overlay mode


thethorne48:
## `heddle context set` orphaned 407 commits of upstream history

Found by an independent evaluator doing a cold usability run on a real repository, and **verified
directly afterwards**. This is data-loss class: a *metadata* command silently rewrote branch history.

### The repository

BurntSushi/xsv, cloned fresh from GitHub. 407 upstream commits.

### Verified state after the session

```console
$ git log --oneline master
5c59868 cmd/sample: honor --seed in the indexed random-access path
530f4ff Bootstrap git-overlay before resolving context state

$ git rev-list --count master
2
$ git rev-list --count origin/master
407
$ git merge-base master origin/master
                      # ← no output. NO COMMON ANCESTOR.
```

`master` and `origin/master` are now **completely disconnected lineages**. The 407 upstream commits
survive only via `origin/master` and the reflog.

### The commit that did it

```console
$ git log -1 --format='%H parents:[%P] %s' 530f4ff
530f4ffe072800b941302925872861e83213cf47 parents:[] Bootstrap git-overlay before resolving context state
```

- **`parents: []`** — a root commit.
- 73 files, a full re-snapshot of the working tree.
- It swept in **untracked build output**: `build.log` (+736 lines) and a cargo-touched `Cargo.lock`
  (+517). Neither belongs in history at all.

### How `master` moved

```console
$ git reflog master
5c59868 master@{0}: heddle: write-through current thread
f430466 master@{1}: clone: from https://github.com/BurntSushi/xsv.git
```

Clone put `master` at `f430466` with all 407 commits. One `heddle: write-through current thread`
later, `master` points into the orphan lineage. There is no intermediate entry — the write-through
replaced the ancestry wholesale.

### Sequence the evaluator ran

1. `heddle clone` the repo (history intact — they read 2015-era commits through `heddle log`).
2. `heddle start` a thread, edit one file.
3. **`heddle context set --scope symbol:…`** — annotate a symbol. This created `530f4ff`.
4. `heddle ready`, `heddle land` — write-through moved `master` onto the orphan root.

Step 3 is a metadata write. It should not produce a git commit at all, let alone a root one.

### Why this is the highest-severity finding we have

- **It contradicts an explicit promise.** `heddle init` prints *"Git commits stay in Git storage,
  worktree stays clean."* The evaluator quoted that back when reporting this.
[...]

PR #1151, opened 2026-07-30
fix(client): close hosted Iroh endpoints

### LINKED ISSUE #1143: fix(client): successful push prints ERROR — Iroh endpoint is never closed
## Problem

A **successful** push prints an ERROR line. Observed on `heddle` 0.10.4 at `091d5ca`, pushing to a local weft over the native lane:

```text
[ok] connected to 127.0.0.1:8421
[ok] pushed to main
remote state: hs-arszx498f2yqwvw
2026-07-30T05:36:28.385771Z ERROR iroh::socket: Endpoint dropped without calling `Endpoint::close`. Aborting ungracefully.
push_rc=0
```

Exit code is 0. The push worked. But the last thing the user sees is `ERROR` and `Aborting ungracefully`, so the natural reading is that something broke — and the natural next step is to check whether the push really landed.

This is a trust problem, not a cosmetic one. A tool that prints `ERROR` on success teaches people to ignore its errors.

## Cause

`crates/client/src/hosted/connection.rs`:

```rust
pub(super) struct HostedConnection {
    pub(super) _endpoint: Endpoint,
    pub(super) connection: iroh::endpoint::Connection,
}

impl Drop for HostedConnection {
    fn drop(&mut self) {
        self.connection.close(0u32.into(), b"Heddle client closed");
    }
}
```

`Drop` closes the **connection** and never the **endpoint**. Iroh requires an explicit `Endpoint::close()`; dropping without it emits that error from `iroh::socket`.

The `_endpoint` underscore is the tell: the field exists to keep the endpoint alive for the connection's lifetime, and closing it was never wired up.

## The design decision this needs

**`Endpoint::close()` is async. `Drop` is not.** That is presumably why this was punted, and it is the real problem to solve rather than route around. Pick deliberately:

- **Explicit async teardown** — give `HostedConnection` an `async fn close()` and call it at every site that finishes with a connection. Correct and predictable; costs an audit of every call path, and anything that returns early or errors out must still close.
- **Spawn from `Drop`** — detach a task to close the endpoint. Cheap, but `Drop` can run during shutdown when the runtime is already gone, so it will sometimes silently not run and sometimes panic. Do not choose this without saying what happens on the shutdown path.
- **Block in `Drop`** — `block_in_place` plus `Handle::block_on`. Only valid on a multi-thread runtime, and blocking in `Drop` is its own hazard.

State which you chose and why, including what it does on the error and early-return paths, not just the happy one.

## Verification

- Push, pull, and clone over the native lane each complete with **no `iroh::socket` ERROR line**, and still exit 0. Show the actual output of each.
- **Show the error present before your change and absent after**, same command. This reproduces trivially, so a green-only run is not evidence.
- Exercise a **failing** operation too — a push to an unreachable or unauthorised remote. The endpoint must still be closed on the error path, and the real error must still be reported. A fix that only closes on success has moved the bug rather than fixed it.
- Confirm no new hang or added latency on exit. Trading a spurious error for a two-second stall at the end of every push is not an improvement — time it.

## Notes

- `crates/client/src/hosted/call.rs:435` and `:462` also construct endpoints; check whether they have the same defect or are test-only, and say which.
- Do not silence the message by lowering the log level or filtering `iroh::socket`. The message is correct — we are genuinely leaking an endpoint. Suppressing it would hide a real resource leak and leave the leak in place.
- Found while pre-flighting a cold usability study; four evaluators are exercising this exact path, so expect it in their reports too.

forge-zephyr:
## Summary

- Add explicit async teardown for `HostedClient`: close the QUIC connection, then await `Endpoint::close()`.
- Run teardown after every owned hosted-client operation on both success and error paths, including push, pull, clone, auth, whoami, support, approvals, and lazy hydration shutdown.
- Close endpoints when connection establishment fails and avoid cancelling the lazy hydration dial after endpoint binding.
- Keep the real operation error as the returned error; no log filtering or transport/auth changes.

## Closes

Closes #1143

### Same successful push before

```text
[ok] connected to 127.0.0.1:8421
[ok] created hosted spool issue-1143-codex/repro-source
Next: heddle remote add origin heddle://127.0.0.1:8421/issue-1143-codex/repro-source
[ok] pushed to main
remote state: hs-cg80gnfeegt5ca2
2026-07-30T05:57:40.433482Z ERROR iroh::socket: Endpoint dropped without calling `Endpoint::close`. Aborting ungracefully.
elapsed=0.27 s
push_rc=0
```

### Same successful push after

```text
[ok] connected to 127.0.0.1:8421
[ok] using hosted spool issue-1143-codex/repro-source
Next: heddle remote add origin heddle://127.0.0.1:8421/issue-1143-codex/repro-source
[ok] pushed to main
remote state: hs-cg80gnfeegt5ca2
elapsed=1.14 s
push_rc=0
```

The graceful QUIC drain adds 0.87 s in this local debug run (0.27 s before, 1.14 s after). That wait is the awaited Iroh endpoint drain; teardown is explicit rather than spawned or blocked from `Drop`.

### Native pull and clone

```text
[ok] connected to 127.0.0.1:8421
[ok] already up to date with main; repository verification checked below
elapsed=1.12 s
pull_rc=0
```

```text
Connected to 127.0.0.1:8421
[ok] cloned heddle://127.0.0.1:8421/issue-1143-codex/repro-source into /home/scratch/issue-1143/repro-clone-final-2
  state: hs-cg80gnfeegt5ca2
elapsed=1.52 s
clone_rc=0
```

### Unauthorized push error path

```text
[ok] connected to 127.0.0.1:8421
Error: remote failure (PermissionDenied): access denied
Next: heddle status
Also: heddle help push
Command exited with non-zero status 77
elapsed=1.09 s
push_rc=77
```

No `iroh::socket` error appeared in any after transcript, while the actual authorization error and exit status remained intact.

### Automated checks

- `cargo build --locked -p heddle-client -p heddle-cli --tests --bin heddle --bin heddle-fuse-worker`
- `cargo clippy --locked -p heddle-client -p heddle-cli --lib --bins --tests --examples -- -D warnings`
- `cargo test --locked -p heddle-client` — 139 unit, 13 contract, and 1 inertness test passed
- `cargo check --locked -p heddle-cli --no-default-features --features git-overlay,client,semantic,zstd`
- `cargo check --locked -p heddle-cli --no-default-features --features native,semantic,zstd`

The combined CLI integration suite cannot use the mandated `TMPDIR=/home/scratch` cleanly on this evaluation host because `/home/scratch/.git` is a pre-existing read-only parent repository; a one-test isolation repro exits 74 with `Read-only file system` during `heddle init`, before the patch path executes. The CI runner does not have that parent-repository collision.

---
<a href="https://app.blacksmith.sh/HeddleCo/codesmith/heddle/pr/1151"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1787985008&installation_model_id=21489&pr_number=1151&repository=HeddleCo%2Fheddle&return_to=https%3A%2F%2Fgithub.com%2FHeddleCo%2Fheddle%2Fpull%2F1151&signature=a284798a7f68039e9e41cc53d356d02c1595c1904591df045c3f40c664ed7c20"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/crates/cli/src/cli/commands/clone.rs
+++ b/crates/cli/src/cli/commands/clone.rs
@@ -1506,18 +1506,6 @@ async fn clone_network(
         config::UserConfig,
     };
 
-    let CloneOptions {
-        thread,
-        depth,
-        lazy,
-        filter,
-        insecure,
-    } = options;
-    let depth = *depth;
-    // `--filter blob:none` is a synonym for `--lazy` on hosted/network
-    // remotes; both produce a clone whose blob content is hydrated on demand.
-    let lazy = *lazy || filter.is_some();
-
     let user_config = UserConfig::load_default()?;
     // On every network-connecting command, TLS/auth config validation
     // (`heddle_client_config`) must succeed before any irreversible
@@ -1526,11 +1514,46 @@ async fn clone_network(
     // no partial on-disk artifact.
     let session =
         HostedSession::build(&user_config, server_key, HostedAuthMode::CredentialFallback)?
-            .with_allow_insecure(*insecure);
+            .with_allow_insecure(options.insecure);
     let repo_path = repo_path.context("network remotes must include a hosted repository path")?;
 
-    let json_output = should_output_json(cli, None);
     let mut client = session.connect(addr).await?;
+    let result = clone_network_connected(
+        cli,
+        addr,
+        repo_path,
+        local_path,
+        options,
+        endpoint_spec,
+        &mut client,
+    )
+    .await;
+    client.close().await;
+    result
+}
+
+#[cfg(feature = "client")]
+async fn clone_network_connected(
+    cli: &Cli,
+    addr: std::net::SocketAddr,
+    repo_path: &str,
+    local_path: &Path,
+    options: &CloneOptions,
+    endpoint_spec: String,
+    client: &mut HostedClient,
+) -> Result<()> {
+    let CloneOptions {
+        thread,
+        depth,
+        lazy,
+        filter,
+        insecure: _,
+    } = options;
+    let depth = *depth;
+    // `--filter blob:none` is a synonym for `--lazy` on hosted/network
+    // remotes; both produce a clone whose blob content is hydrated on demand.
+    let lazy = *lazy || filter.is_some();
+    let json_output = should_output_json(cli, None);
 
     if json_output {
         println!(
@@ -1630,8 +1653,7 @@ async fn clone_network(
         // hosted CollaborationService discussions for the cloned head into the
         // local op-log so `discuss list` / `discuss show` see them. Best-effort:
         // a fetch hiccup warns rather than failing an otherwise-good clone.
-        match crate::client::discussion_sync::pull_discussions(&local_repo, &mut client, repo_path)
-            .await
+        match crate::client::discussion_sync::pull_discussions(&local_repo, client, repo_path).await
         {
             Ok(_) => {}
             Err(error) => {
@@ -1644,7 +1666,7 @@ async fn clone_network(
         // Read path for hosted context annotations (heddle context): materialize
         // the hosted head's annotations into the local Context attachment so
         // `context list` sees them. Best-effort, mirroring discussions.
-        match crate::client::context_sync::pull_context(&local_repo, &mut client, repo_path).await {
+        match crate::client::context_sync::pull_context(&local_repo, client, repo_path).await {
             Ok(_) => {}
             Err(error) => {
                 eprintln!("{} context sync skipped: {error:#}", style::warn_marker());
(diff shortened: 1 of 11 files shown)


thethorne48:
Merging. The design choice is the right one and the evidence is complete: explicit async teardown rather than spawn-from-`Drop` or block-in-`Drop`, applied on **both success and error paths** across push, pull, clone, auth, whoami, support, approvals, and lazy-hydration shutdown — and endpoints closed when connection establishment itself fails, which is the case most likely to be missed.

Two things I checked specifically because they are how this kind of fix usually goes wrong:

- **The real error still surfaces.** The failing-path run returns `remote failure (PermissionDenied): access denied` with `push_rc=77`, not a teardown error masking it.
- **No log filtering.** The message was accurate — we were genuinely leaking an endpoint — and suppressing it would have hidden a real resource leak.

## On the 0.87 s

Thank you for measuring and stating it plainly instead of omitting it. That is the disclosure that lets me make a decision rather than discover it later.

I am accepting it, with a follow-up, on these grounds:

- It is a **local debug build**. Debug Iroh is materially slower than release, so 0.87 s is an upper bound, not the shipped cost.
- The alternative is a spurious `ERROR ... Aborting ungracefully` on every successful push, which teaches users to ignore the tool's errors. That is a worse and less reversible cost than latency.
- A graceful QUIC drain is the correct behaviour, not overhead we invented.

But it is not free, and for an agent-native tool doing many small operations a per-operation second matters. Filing a follow-up to **measure it in release and bound the drain if it is still significant** — a close with a deadline, falling back to abort after it, would keep the fix while capping the tail.

Not asking you to do that here; this PR is correct as it stands.

## Provenance

Found while pre-flighting a cold usability study — a successful push printing `ERROR` was the first thing I saw. Four independent evaluators then hit it too. It is one of thirteen issues from that run (HeddleCo/heddle#1144 through #1156), and one of the more straightforward ones to have landed cleanly.

Issue #1147, opened 2026-07-30
[p0] ancestor-walking discovery writes outside the repo and bricks repos it does not own


thethorne48:
## Two symptoms, one root cause

Repository discovery walks *up* the filesystem looking for `.git`, and treats the nearest ancestor as the repo root — even when the current directory has its own `.heddle`. That single design choice produces both failures below. Found in a cold usability run; both verified afterwards.

---

## Symptom 1: `heddle init` creates a `.git` outside the repository

Running `heddle init probe` in a fresh directory under a shared scratch root classified it as *git-overlay* rooted three levels up and **created `/home/scratch/.git/info/exclude`**:

```console
$ ls /home/scratch/.git
info
$ cat /home/scratch/.git/info/exclude
# Heddle local metadata
.heddle/
```

Verified present, timestamped to the evaluator's session. Note what it is: no `HEAD`, no `objects`, no `refs`. **`heddle init` minted a fake `.git` on a shared 350 GB scratch root** used concurrently by other processes, as a side effect of initialising an unrelated project.

This is not hypothetical damage. That marker breaks Heddle's **own test suite** — heddle's tests create repos under `TMPDIR` and walk up looking for `.git`/`.heddle`; when they find one, hundreds of tests fail for reasons unrelated to the change under test. This exact directory has now poisoned that root twice.

A VCS must never write to a directory above the repository it was asked to initialise. There is no configuration under which that is correct.

## Symptom 2: an unreadable ancestor bricks every command, with no diagnosable error

After the above, *every* heddle command in *every* repo under that root began failing:

```console
$ heddle status
Error: configuration error: io error: Permission denied (os error 13)
Next: heddle status
```

No path. No hint. The `Next:` advice is the command that just failed. `doctor` fails with the same error it exists to diagnose. `verify` and `fsck` reported healthy.

The evaluator diagnosed it only with **`strace`**, finding heddle recursively crawling the entire shared root — roughly half a million syscalls through other users' directories — and aborting on the first unreadable one. They confirmed the mechanism by planting a `git init` fence between their repo and the root: command latency went from *7 seconds then error* to *0.1 s*.

Their repo, with five states of real history, was recoverable only by giving it an empty `.git` of its own. It now carries an annotation warning future readers never to delete that empty directory. That sentence is the finding.

## Why this is P0

- **It writes to filesystem locations the user did not name.** Worst class of side effect a tool can have.
[...]

PR #1157, opened 2026-07-30
fix(cli): isolate HEDDLE_HOME and unify principal resolution

### LINKED ISSUE #1146: fix(cli): HEDDLE_HOME is silently ignored — global state leaks to the shared home
## Problem

`HEDDLE_HOME` has no effect. It is referenced ~54 times in the source, so it is clearly intended to work, but setting it does not relocate the home directory and produces no error or warning.

## Verified

Set **inline in the same command**, so this is not a shell/export artifact:

```console
$ HEDDLE_HOME=/home/scratch/hometest/myhome \
  HEDDLE_CONFIG=/home/scratch/hometest/my.toml \
  heddle init --principal-name T --principal-email t@x.io

$ ls /home/scratch/hometest
my.toml   repo          # ← my.toml created; myhome/ never created
```

**`HEDDLE_CONFIG` is honoured. `HEDDLE_HOME` is not.** That asymmetry is the clearest statement of the bug: one of the pair works, so a user reasonably assumes both do.

## Impact observed

Four independent evaluators ran concurrently on one machine, each instructed to set `HEDDLE_HOME` to its own directory for isolation. **None of the four directories was created.** All four shared `~/.heddle` and `~/.config/heddle`, and one evaluator's `heddle auth login` wrote device and session state into the shared home — they flagged it themselves, unprompted:

> my `heddle auth login` attempt wrote session/device state to the shared `~/.heddle` because `HEDDLE_HOME` was ignored

So the failure mode is: **silent loss of isolation, with credential-bearing state going somewhere the caller did not choose.** On a shared or multi-agent machine that is a correctness and a privacy problem, not an ergonomics one. It also means any test harness relying on `HEDDLE_HOME` for isolation has been running without it — including ours.

## Acceptance criteria

- [ ] `HEDDLE_HOME=<path>` relocates everything that would otherwise land in `~/.heddle`, including credentials, device identity, and session state.
- [ ] **Demonstrated failing first:** show the current behaviour (dir not created, state in `~/.heddle`), then show it fixed with the same command. This reproduces in one command, so a green-only run is not evidence.
- [ ] Two concurrent invocations with different `HEDDLE_HOME` values do not share state — prove it, do not assert it.
- [ ] If any path legitimately cannot be relocated, it **fails loudly** rather than silently falling back to the shared home.
- [ ] `heddle help advanced` documents `HEDDLE_HOME` and `HEDDLE_CONFIG` together. Neither appears in help output today.

## Notes

- Check `HEDDLE_CONFIG`'s implementation for the pattern that works and whether `HEDDLE_HOME` is simply never read on the resolution path, versus read and then discarded.
- Silent fallback is the specific thing to remove. A user who sets the variable and gets no error has every reason to believe they are isolated.

### LINKED ISSUE #1150: identity: init, status, and capture disagree about who the principal is
## Problem

Four surfaces give four different answers to "who am I", and the one that wins is a stale value from a shared global config — so captures get attributed to the wrong person silently.

## Observed

```console
$ heddle init --principal-name 'Luke Thorne' --principal-email <addr>
Principal: Luke Thorne <...> from user_config
... updated default principal attribution

$ heddle status
no principal configured; first capture would use Unknown

$ heddle capture -m "..."
# attributed to: Coldrun Agent <agent@example.invalid>
```

- `init` says it set the principal and reports the value it was given.
- `status` says none is configured and threatens `Unknown`.
- `capture` uses **neither** — it picks up a stale principal from the shared `~/.config/heddle/config.toml`.
- `whoami` is a fourth answer: server-auth identity only, unrelated to capture attribution.

`HEDDLE_PRINCIPAL_NAME` / `HEDDLE_PRINCIPAL_EMAIL` do work, which is how the evaluator eventually forced correct attribution.

## Why it matters

Attribution is provenance, and provenance is one of the things this tool sells. A capture stamped with a stranger's name — silently, after `init` reported success — is a correctness bug, not a cosmetic one. The evaluator's attempt to fix it retroactively is what led them into HeddleCo/heddle#1148 (the destructive `undo` and its recovery deadlock), so this defect also has a nasty downstream.

It is made worse by HeddleCo/heddle#1146: `HEDDLE_HOME` is ignored, so the "shared global config" is shared across every concurrent user on the machine. Four evaluators ran at once here; whoever wrote that config last would have attributed everyone else's work.

## Acceptance criteria

- [ ] One resolution order for principal identity, documented, with **`init`, `status`, and `capture` all reporting the same answer**. Show all three agreeing after the fix.
- [ ] `init --principal-name/--principal-email` actually persists what it claims to persist — or says it did not.
- [ ] `status` never says "no principal configured" when one is resolvable, and never threatens `Unknown` when a fallback exists.
- [ ] When identity comes from a shared global config rather than the repo or flags, **say so** — the source should be visible, since that is the case most likely to be wrong.
- [ ] Reproduce the three-way disagreement first and post it, then show it resolved.

## Notes

- `init` reporting `from user_config` while `status` disagrees suggests the two read different resolution paths. Start there.
- Fix HeddleCo/heddle#1146 alongside or before this — while `HEDDLE_HOME` is ignored, per-user isolation of the config that wins is impossible, and any fix here is testable only serially.
- Same cold run: HeddleCo/heddle#1144, #1145, #1146, #1147, #1148, #1149.

forge-zephyr:
## Summary

- #1146: make `HEDDLE_HOME` the root for the default user config as well as credentials, device identity, persisted operation IDs, and tip-session state. `HEDDLE_CONFIG` remains the higher-precedence file override. `init` now creates an explicitly requested home even when `HEDDLE_CONFIG` is also set, and fails with the requested path instead of silently falling back when that home is unusable.
- #1150: centralize principal resolution as environment → repository config → Git config → user config → Unknown, inject user config into `status`, and expose the selected source in init/status/capture text plus capture JSON. User-config attribution is labeled as shared global config.
- Document `HEDDLE_HOME`, `HEDDLE_CONFIG`, and principal precedence in `heddle help advanced`.

## Closes

Closes #1146

Closes #1150

## Test evidence

Failing first on the untouched release binary (#1146):

```text
Principal: T <t@x.io> from user_config
$ find .../before-1146-hermetic -maxdepth 1 -mindepth 1 -printf '%f\n'
my.toml
repo
shared-home
# myhome is absent
```

Failing first with two evaluator homes sharing one `HOME` (#1150): evaluator A initialized Luke, evaluator B then initialized Coldrun Agent, and evaluator A observed:

```text
Identity: no principal configured; the first capture would use Unknown <unknown@example.com>...
Captured state hs-mvhm79b3wy8jxek (a6e343a5)
Saved by: Coldrun Agent <agent@example.invalid>
```

After (#1146, same inline `HEDDLE_HOME` + `HEDDLE_CONFIG` shape):

```text
Principal: T <t@x.io> from user_config (shared global config)
my.toml
myhome
repo
shared-home
```

After concurrent isolation, both inits ran simultaneously and produced separate files while the shared `HOME` remained empty:

```text
heddle-a/config.toml: name = "Luke Thorne"
heddle-a/config.toml: email = "luke@example.test"
heddle-b/config.toml: name = "Coldrun Agent"
heddle-b/config.toml: email = "agent@example.invalid"
```

After (#1150), all three surfaces agree:

```text
Principal: Luke Thorne <luke@example.test> from user_config (shared global config)
Identity: Luke Thorne <luke@example.test> from user_config (shared global config)
Saved by: Luke Thorne <luke@example.test> from user_config (shared global config)
```

Verification:

- `TMPDIR=/home/scratch cargo test -p heddle-cli --test cli_integration identity_resolution` — 4 passed, including concurrent config/session isolation and unusable-home fail-loud behavior.
- `TMPDIR=/home/scratch cargo test -p heddle-cli --lib` — 452 passed.
- affected cli-contract/cli-render/cli-shared/core library run — 592 passed; 2 unrelated core tempdir tests hit the documented read-only ancestor `.git` discovery hazard from #1147.
- `TMPDIR=/home/scratch cargo test -p heddle-client credentials_path_honors_heddle_home` — passed.
- `TMPDIR=/home/scratch cargo test -p heddle-repo repository_signing --lib` — 10 passed.
- `TMPDIR=/home/scratch cargo build -p heddle-cli` — passed.
- `TMPDIR=/home/scratch cargo clippy -p heddle-cli -p heddle-cli-shared -p heddle-cli-contract -p heddle-cli-render -p heddle-core -p heddle-repo --all-targets -- -D warnings` — passed.
- `rustfmt +nightly --edition 2024 --check <touched Rust files>` — passed; no `cargo fmt` or `cargo clean` used.

No Heddle-owned path remains unable to relocate: credentials, device identity, default user config, operation-ID persistence, and tip-session state all resolve through `HEDDLE_HOME`. External Git/harness/integration homes remain `HOME`-based because they are not Heddle-owned state.

---
<a href="https://app.blacksmith.sh/HeddleCo/codesmith/heddle/pr/1157"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1787987193&installation_model_id=21489&pr_number=1157&repository=HeddleCo%2Fheddle&return_to=https%3A%2F%2Fgithub.com%2FHeddleCo%2Fheddle%2Fpull%2F1157&signature=a1bb43506811e7462e73f6bce00802bf3e32d476caeaa613ddf6553eff01f3d9"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/clients/npm/generated/heddle-schemas.json
+++ b/clients/npm/generated/heddle-schemas.json
@@ -977,6 +977,9 @@
         "principal": {
           "$ref": "#/$defs/CommitPrincipalSchema"
         },
+        "principal_source": {
+          "type": "string"
+        },
         "promotion_suggested": {
           "type": "boolean"
         },
@@ -1024,6 +1027,7 @@
         "state_id",
         "content_hash",
         "principal",
+        "principal_source",
         "promotion_suggested",
         "heavy_impact_paths",
         "signed",
@@ -12101,6 +12105,9 @@
         "principal": {
           "$ref": "#/$defs/CommitPrincipalSchema"
         },
+        "principal_source": {
+          "type": "string"
+        },
         "promotion_suggested": {
           "type": "boolean"
         },
@@ -12148,6 +12155,7 @@
         "state_id",
         "content_hash",
         "principal",
+        "principal_source",
         "promotion_suggested",
         "heavy_impact_paths",
         "signed",
--- a/clients/npm/generated/heddle-schemas.ts
+++ b/clients/npm/generated/heddle-schemas.ts
@@ -125,6 +125,7 @@ export interface AgentCaptureSchema {
   operation_record?: { command: string; idempotency_status: string; op_id: string; replayed: boolean; } | null;
   output_kind: "capture";
   principal: CommitPrincipalSchema;
+  principal_source: string;
   promotion_suggested: boolean;
   recommended_action?: string | null;
   recommended_action_template?: ActionTemplateSchema | null;
@@ -515,6 +516,7 @@ export interface CaptureSchema {
   operation_record?: { command: string; idempotency_status: string; op_id: string; replayed: boolean; } | null;
   output_kind: "capture";
   principal: CommitPrincipalSchema;
+  principal_source: string;
   promotion_suggested: boolean;
   recommended_action?: string | null;
   recommended_action_template?: ActionTemplateSchema | null;
--- a/crates/cli-contract/src/cli/commands/schemas.rs
+++ b/crates/cli-contract/src/cli/commands/schemas.rs
@@ -1223,6 +1223,7 @@ pub struct CaptureSchema {
     pub confidence: Option<f32>,
     pub task_assignment_id: Option<String>,
     pub principal: CommitPrincipalSchema,
+    pub principal_source: String,
     pub agent: Option<CommitAgentSchema>,
     pub promotion_suggested: bool,
     pub heavy_impact_paths: Vec<String>,
--- a/crates/cli-contract/src/cli/help.rs
+++ b/crates/cli-contract/src/cli/help.rs
@@ -520,7 +520,17 @@ docs.\n\
 \n\
 This is intentional. The everyday surface stays minimal so first-time users aren't\n\
 overwhelmed; agents and power users reach for the advanced affordances when they\n\
-need them.\n";
+need them.\n\
+\n\
+Configuration environment:\n\
+  HEDDLE_HOME    Relocates global Heddle state (credentials, device identity,\n\
+                 session state, and the default user config).\n\
+  HEDDLE_CONFIG  Overrides the user config file. Takes precedence over\n\
+                 HEDDLE_HOME; otherwise HEDDLE_HOME uses <path>/config.toml.\n\
+\n\
+Principal resolution (highest first): HEDDLE_PRINCIPAL_NAME and\n\
+HEDDLE_PRINCIPAL_EMAIL, repository config, Git config, user config, then\n\
+Unknown <unknown@example.com>. Init, status, and capture use this same order.\n";
 
 // The single full statement of the --output machine contract (plus the
 // one-paragraph version on the top-level `heddle help`). The global
@@ -1362,6 +1372,17 @@ mod tests {
             !advanced.contains("Advanced commands:"),
             "the flat list header is replaced by area groups: {advanced}"
         );
+    }
+
+    #[test]
+    fn advanced_help_documents_home_and_config_overrides() {
+        use clap::CommandFactory;
+        let cmd = crate::cli::cli_args::Cli::command();
+        let advanced = render_help(&cmd, &["advanced".to_string()]);
+        assert!(advanced.contains("HEDDLE_HOME"));
+        assert!(advanced.contains("HEDDLE_CONFIG"));
+        assert!(advanced.contains("<path>/config.toml"));
+        assert!(advanced.contains("Principal resolution (highest first)"));
     }
 
     /// heddle#652. The grouped advanced surface is exhaustive and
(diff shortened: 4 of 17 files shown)


forge-zephyr:
🤖 Starting. Worktree `workspace/issue-1157`, branch `task/1157-fix-cli-isolate-heddle-home-and-unify-pr`. File scope: (inferred from epic; declare in PR).

forge-zephyr:
♻️ Released by orchestrator. Reason: manual release. Worktrees preserved for inspection: `workspace/issue-1157`.

PR #1158, opened 2026-07-30
fix(discovery): keep repository resolution inside root

### LINKED ISSUE #1144: [p0] data loss: heddle context set orphans existing git history in overlay mode

### LINKED ISSUE #1147: [p0] ancestor-walking discovery writes outside the repo and bricks repos it does not own

forge-zephyr:
## Summary

- Make cwd Heddle metadata authoritative before considering any ancestor Git repository.
- Restrict init and Git-overlay metadata writes to a valid `.git` rooted at the resolved repository root.
- Bound Heddle ancestor discovery at filesystem boundaries by default, with the Git-compatible `GIT_DISCOVERY_ACROSS_FILESYSTEM` opt-out, and treat inaccessible marker probes as misses.
- Route plain-Git fallback through environment-aware Sley discovery and add paths to discovery and overlay metadata I/O failures.

## Closes

Closes #1147.

Does not close #1144. The preserved failure does not reproduce on current main or this branch, but this change does not alter the context bootstrap/write-through mechanism that produced its orphan root; #1144 should remain open independently.

## Test evidence

Before the fix, initializing `outer/fresh/probe` beneath `outer/.git` reported `repository_mode: git-overlay`, changed the ancestor `.git/info/exclude` SHA from `6671fe...` to `d23a58...`, and appended the Heddle metadata marker. Native status beneath that Git worktree exited 78 on an unreadable sibling with `configuration error: io error: Permission denied (os error 13)` and no path.

After the fix, the same init reports `repository_mode: native-heddle`, `git_detected: false`, and leaves the ancestor exclude SHA unchanged. The same native status exits 0 and reports `native-heddle` / `heddle-native`; a file trace contains no access to the ancestor `.git` or unreadable sibling.

Controlled timing on the compact fixture was 0.14s before and 0.17s after for init, and 0.02s failing before versus 0.02s successful after for status. Status tracing fell from 1,000 syscalls to 917, and the ancestor scan disappeared entirely. The evaluator separately measured the pathological shared-root case at about 7s before versus 0.1s with a Git boundary fence.

Regression coverage asserts: no file outside the requested init root changes; cwd native metadata wins over a readable ancestor Git worktree; unreadable ancestor-worktree entries are non-fatal; local metadata I/O errors name the failing path; and discovery stops at a filesystem boundary unless explicitly opted out.

Verification:
- `cargo test --locked -p heddle-repo --lib`: 605 passed, 2 ignored
- `cargo test --locked -p heddle-core --lib`: 413 passed
- `cargo test --locked -p heddle-cli --lib`: 452 passed
- `cargo test --locked -p heddle-cli --test onboarding_contract`: 10 passed
- `cargo test --locked -p heddle-cli --tests --no-fail-fast`: all test binaries passed; primary integration binary 867 passed, 19 ignored
- touched files formatted with `rustfmt +nightly --edition 2024`

For #1144, read-only inspection confirmed the preserved orphan root (`530f4ff...`, no parents) and 407-commit upstream history. Disposable `init` and exact `clone -> start -> context set -> capture -> ready -> land` reproductions both retained the real Git tip as parent and a valid merge base. That means #1144 does not reproduce here, not that this PR fixes its bootstrap path.

---
<a href="https://app.blacksmith.sh/HeddleCo/codesmith/heddle/pr/1158"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1787987275&installation_model_id=21489&pr_number=1158&repository=HeddleCo%2Fheddle&return_to=https%3A%2F%2Fgithub.com%2FHeddleCo%2Fheddle%2Fpull%2F1158&signature=c2da4571cd4b4f7fe252d185ea9cd6874677710c10e20ce672c1fca01cd260c0"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/crates/cli/src/cli/commands/init.rs
+++ b/crates/cli/src/cli/commands/init.rs
@@ -10,9 +10,8 @@ use heddle_core::{
     select_init_principal,
 };
 use objects::object::Principal;
-use repo::{Repository, RepositoryCapability};
+use repo::{Repository, RepositoryCapability, open_git_repository_at_root};
 use serde::Serialize;
-use sley::Repository as SleyRepository;
 use tracing::{debug, info};
 
 use super::{
@@ -103,7 +102,10 @@ pub fn cmd_init(cli: &Cli, args: InitArgs) -> Result<()> {
         None
     };
 
-    let git = SleyRepository::discover(&path).ok();
+    // `init <path>` owns exactly that path. An ancestor Git worktree is not
+    // the requested repository and must never redirect initialization or
+    // receive exclude-file writes.
+    let git = open_git_repository_at_root(&path)?;
     let existing_repo = if path.join(".heddle").exists() {
         Some(Repository::open(&path)?)
     } else {
--- a/crates/cli/src/cli/commands/status.rs
+++ b/crates/cli/src/cli/commands/status.rs
@@ -19,6 +19,7 @@ use heddle_core::{
 };
 use repo::{
     RepoConfig, Repository, ThreadFreshness, ThreadMode, ThreadState, WorktreeCompareProfile,
+    discover_heddle_root,
 };
 #[cfg(feature = "client")]
 use serde::Deserialize;
@@ -131,7 +132,10 @@ fn try_fast_short_status_report(cli: &Cli) -> Result<Option<FastShortStatusRepor
 
 fn pending_incomplete_land_marker_exists(start: &Path) -> bool {
     let canonical = start.canonicalize().unwrap_or_else(|_| start.to_path_buf());
-    if let Ok(git) = SleyRepository::discover(&canonical)
+    if let Some(heddle_root) = discover_heddle_root(&canonical) {
+        return heddle_root.join(".heddle/incomplete-land.json").is_file();
+    }
+    if let Ok(git) = SleyRepository::open_from_environment(&canonical)
         && let Some(workdir) = git.workdir()
     {
         // A nested plain-Git repository under a Heddle checkout is its own
@@ -144,7 +148,15 @@ fn pending_incomplete_land_marker_exists(start: &Path) -> bool {
 }
 
 fn fast_short_repo_config(start: &Path) -> Result<Option<RepoConfig>> {
-    let Ok(git) = SleyRepository::discover(start) else {
+    if let Some(heddle_root) = discover_heddle_root(start) {
+        let config_path = heddle_root.join(".heddle/config.toml");
+        return if config_path.is_file() {
+            Ok(Some(RepoConfig::load_for_repository(&config_path)?))
+        } else {
+            Ok(None)
+        };
+    }
+    let Ok(git) = SleyRepository::open_from_environment(start) else {
         return Ok(None);
     };
     let Some(workdir) = git.workdir() else {
--- a/crates/cli/src/cli/commands/verify.rs
+++ b/crates/cli/src/cli/commands/verify.rs
@@ -8,7 +8,7 @@ use heddle_core::{
     MachineContractInput, RepositorySetupGuidance, RepositoryVerificationState, VerificationCheck,
     VerifyOptions, VerifyReport, repository_setup_guidance, verify as core_verify,
 };
-use repo::Repository;
+use repo::{Repository, discover_heddle_root};
 
 use super::{RecoveryAdvice, action_line::print_next};
 use crate::{
@@ -142,17 +142,10 @@ fn verify_execution_context_from_cli(cli: &Cli, start: &Path) -> Result<VerifyEx
 /// True when `start` or an ancestor already has a Heddle sidecar (main repo or
 /// worktree pointer). Used to avoid auto-bootstrap on observe-only verify.
 fn heddle_sidecar_present(start: &Path) -> bool {
-    let mut current = Some(start);
-    while let Some(dir) = current {
-        let heddle = dir.join(".heddle");
-        if heddle.is_dir()
-            && (heddle.join("objects").is_dir() || heddle.join("objectstore").is_file())
-        {
-            return true;
-        }
-        current = dir.parent();
-    }
-    false
+    discover_heddle_root(start).is_some_and(|root| {
+        let heddle = root.join(".heddle");
+        heddle.join("objects").is_dir() || heddle.join("objectstore").is_file()
+    })
 }
 
 fn render_verify(output: &VerifyReport, verbose: bool, as_json: bool) -> Result<()> {
--- a/crates/cli/tests/onboarding_contract.rs
+++ b/crates/cli/tests/onboarding_contract.rs
@@ -54,6 +54,41 @@ fn json(output: &Output) -> Value {
     serde_json::from_slice(&output.stdout).expect("parse heddle JSON")
 }
 
+fn snapshot_outside_repository(
+    root: &Path,
+    repository: &Path,
+) -> Vec<(std::path::PathBuf, Vec<u8>)> {
+    fn visit(
+        root: &Path,
+        directory: &Path,
+        repository: &Path,
+        snapshot: &mut Vec<(std::path::PathBuf, Vec<u8>)>,
+    ) {
+        let mut entries = fs::read_dir(directory)
+            .unwrap()
+            .map(|entry| entry.unwrap().path())
+            .collect::<Vec<_>>();
+        entries.sort();
+        for path in entries {
+            if path == repository {
+                continue;
+            }
+            if path.is_dir() {
+                visit(root, &path, repository, snapshot);
+            } else if path.is_file() {
+                snapshot.push((
+                    path.strip_prefix(root).unwrap().to_path_buf(),
+                    fs::read(path).unwrap(),
+                ));
+            }
+        }
+    }
+
+    let mut snapshot = Vec::new();
+    visit(root, root, repository, &mut snapshot);
+    snapshot
+}
+
 #[test]
 fn committed_git_and_unborn_git_both_enter_through_init() {
     for committed in [false, true] {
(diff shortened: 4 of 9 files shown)


thethorne48:
Merging. All three invariants land with tests, and the honesty about #1144 is the part I want to call out.

Verified in the report:
- No writes anywhere outside the requested init root.
- A cwd native `.heddle` wins over any readable ancestor `.git`.
- Unreadable ancestor entries are non-fatal.
- Plus filesystem-boundary stopping with `GIT_DISCOVERY_ACROSS_FILESYSTEM`, and I/O errors that name the failing path — which was the thing that made the original bug undiagnosable without `strace`.

Coverage is real: 605 + 413 + 452 library tests, 867 in the primary integration binary.

## On HeddleCo/heddle#1144

You declined to claim credit for fixing it, inspected the preserved evidence read-only, and reported that the exact `clone → start → context set → capture → ready → land` sequence no longer reproduces on either main or this branch — while stating plainly that the context bootstrap/write-through mechanism was neither changed nor tested by this work.

That is exactly right, and it saved a false close on a P0. I independently tried a second hypothesis afterwards — that the poisoned ancestor `.git` was the trigger — and disproved it too: pre-fix binary, planted ancestor, history intact at 5 commits. So #1144 stays open with three failed reproductions recorded on it and the trigger still unidentified.

## Wider than the issue described

Worth knowing what this unblocks: the ancestor-walk was also breaking heddle's **own** test suite inside codex sandboxes, which synthesise a read-only `.git` above the work root. The agent on HeddleCo/heddle#1143 could not run the CLI integration suite for that reason, and HeddleCo/heddle#1150's agent hit the same hazard. This fix should clear both.

Issue #1155, opened 2026-07-30
ignore handling: gitignore not honored, undocumented, and build junk lands in history


thethorne48:
## Problem

Three of four evaluators in a cold usability run had build artifacts swept into permanent history within their first few captures. The ignore story fails in three distinct ways.

### 1. Thread checkouts do not honour the repo's `.gitignore`

After running pytest inside an isolated thread, `heddle status` listed **19 "added" files** — every `.pyc` and `.pytest_cache` entry — despite `__pycache__/` being in the repo's `.gitignore`. `capture` has no path selection and no staging area, so the first capture swept **14 binary artifacts into permanent history**. Recovery required `undo` → `rm -rf` → recapture.

A second evaluator hit the same thing on a fresh native repo: their first capture silently included `__pycache__/*.pyc`, and a third had `status` wanting to capture a **1,819-file virtualenv**.

### 2. `.heddleignore` is undocumented

There is no `heddle help ignore`. The file is mentioned nowhere in any help page. All three evaluators who needed it **guessed** the name and gitignore-compatible syntax by analogy, and it happened to work:

> I guessed `.heddleignore` by analogy and it worked — undocumented, discovered only by luck

`heddle init` reports `"installed_heddleignore": false` in its JSON, which proves the tool knows the concept exists, yet nothing suggests installing one. A `[worktree] ignore` config key also exists and is equally undocumented.

### 3. No path scoping on capture

`capture` takes everything. With no staging area and no `--path`, the only way to exclude something is to have ignored it beforehand — which, combined with (2), means the first capture in any project with a build step is likely to be wrong.

## Why this compounds

Ignore breakage is what led one evaluator into HeddleCo/heddle#1148: the junk files blocked `undo --recover`, and the `.heddleignore` that would have excluded them had been deleted by the undo. So this defect supplies the fuel for that deadlock. Worth fixing alongside it.

It also undermines a claim the tool makes about itself. Git's ignore handling is invisible because it works; here it is invisible because it is undocumented, and the failure is silent and permanent.

## Acceptance criteria

- [ ] A thread checkout honours the repository's ignore rules — including the parent repo's `.gitignore` in Git-Overlay mode. Reproduce the 19-file case first and post it.
- [ ] `heddle init` either installs a sensible `.heddleignore` or tells the user how to, rather than reporting `installed_heddleignore: false` silently.
[...]

PR #1160, opened 2026-07-30
fix(undo): prevent recovery deadlock

### LINKED ISSUE #1148: [p0] undo deletes uncommitted source files and can deadlock its own recovery
## Problem

`heddle undo` of a capture has `git reset --hard` semantics: it rewinds the worktree and **deletes uncommitted source files from disk**. Worse, the documented recovery path can become unreachable, because the undo also removes the ignore file whose absence then blocks recovery.

Found in a cold usability run by an evaluator who lost their working files and escaped only by reconstructing state by hand.

## The sequence

1. Evaluator ran `heddle undo` to correct a capture's attribution.
2. The worktree rewound. `ls` in the source directory showed **only `__pycache__`** — their source files were gone.
3. Recovery *is* advertised, and the tool said so: `Preserved pre-undo state ... as .undo-recovery / Next: heddle undo --recover`. Good.
4. `heddle undo --recover` **refused**: `1815 unsaved worktree path(s)`.
5. Those 1815 paths were a virtualenv — previously ignored, now unignored, **because the undo had also deleted their `.heddleignore`**.
6. Capturing those paths to clear the block failed: `Error: symlink target escapes repository: python3` (no file path given; it was a venv symlink).

So: capture blocked by a symlink → recover blocked by uncaptured paths → **the file that would unblock both was deleted by the undo itself.** They escaped by hand-recreating `.heddleignore`.

Their assessment: *"A less stubborn user loses their morning's work here, with the recovery mechanism itself telling them no."*

## Why this is P0

- **Silent destruction of uncommitted work.** The hard/soft distinction was never offered and was communicated only after the fact. Git makes you type `--hard`.
- **The recovery mechanism has a deadlock.** A safety net that can be disabled by the operation it protects against is not a safety net.
- **The trigger was ordinary.** Fixing an attribution mistake is a beginner action, not a power move.

## What needs to happen

1. **Decide and state the semantics.** If `undo` must rewind the worktree, it should say so *before* acting and require confirmation or an explicit flag for the destructive form. A soft variant that rewinds the state but leaves files in place is the behaviour most users will expect — the third evaluator noted `undo --recover` restoring edits *as worktree changes* was "ideal semantics", so that shape already exists and arguably should be the default.
2. **`undo` must not delete `.heddleignore`** (or any repo-config file whose absence changes what is considered dirty). If it is captured state, restoring it is part of recovery; if it is not captured, it should be untouched.
3. **Break the deadlock.** `undo --recover` must be able to run when the only blockers are paths it did not create. Recovery must not be gated on a clean worktree it just dirtied.
4. **`symlink target escapes repository` must name the path.** An error about a file that does not say which file is not actionable.

## Verification

- **Reproduce all six steps first and post the output**, including the empty source directory and the `1815 unsaved worktree path(s)` refusal. Do not change anything until the deadlock is reproduced — a fix validated only against the happy path will leave it.
- After the fix: an `undo` following a capture with uncommitted edits present must either preserve them or refuse with a clear statement of what would be lost. Show both.
- Show `undo --recover` succeeding in the exact state that previously blocked it.
- Confirm `.heddleignore` survives an undo.
- Add regression tests: undo-then-recover round-trips with an ignore file present, and with untracked ignored junk in the tree.

## Notes

- The evaluator rated the recovery surface positively when it works — `undo --list` / `--preview` was called "a coherent recovery surface (when it isn't deadlocking)". This is a good feature with a sharp edge, not a bad design.
- Same cold run: HeddleCo/heddle#1144, #1145, #1146, #1147. This one and #1147 are the two that cost the evaluator actual work.

forge-zephyr:
## Summary

- require `--hard` before an undo may rewind worktree files; plain undo refuses pre-mutation and points to `--preview`
- preserve `.heddleignore` during every undo worktree rewind, plus `.gitignore` for Git-overlay repositories
- make recovery overlay-aware: unrelated dirty paths survive, while differing changes at recovery target paths still fail safely
- include the repository-relative symlink path and target in escaping-symlink diagnostics

The chosen semantics keep undo explicit: operations that do not rewrite the worktree remain available as plain `heddle undo`, while tree-rewriting inverses require `heddle undo --hard`. `--hard` still refuses and names unsaved paths. This keeps the destructive boundary visible without changing recovery's existing ideal shape: recovery restores the saved state as worktree changes without moving HEAD.

## Closes

Closes #1148

## Test evidence

Before the fix, the complete six-step deadlock reproduced in a fresh native repository:

```text
$ find .venv \( -type f -o -type l \) | wc -l
1814
$ heddle capture -m initial-source
Captured state hs-fs2vtjsh9wg6p2k (7e45bd4b)
exit=0
$ heddle undo
Undid 1 saved change
Preserved pre-undo state hs-fs2vtjsh9wg6p2k as `.undo-recovery`
Next: heddle undo --recover
exit=0
$ ls -la src
total 12
drwxr-xr-x 3 heddleco heddleco 4096 Jul 30 08:36 .
drwx------ 6 heddleco heddleco 4096 Jul 30 08:36 ..
drwxr-xr-x 2 heddleco heddleco 4096 Jul 30 08:36 __pycache__
$ test -e .heddleignore; echo $?
1
$ heddle undo --recover
Error: Capture worktree changes before recover the pre-undo state
Next: heddle capture -m "..."
Paths: unsaved worktree path(s): .venv/bin/python, .venv/bin/python3, .venv/lib/file-0001, .venv/lib/file-0002, .venv/lib/file-0003, .venv/lib/file-0004, .venv/lib/file-0005, .venv/lib/file-0006, .venv/lib/file-0007, .venv/lib/file-0008, .venv/lib/file-0009, .venv/lib/file-0010, and 1803 more
exit=65
$ heddle capture -m clear-recovery-blocker
Error: symlink target escapes repository: python3
exit=74
```

The two new load-bearing regressions were also run against the pre-fix binary and failed:

```text
test_undo_recover_round_trips_with_heddleignore_present ... FAILED
NotFound: .heddleignore

test_undo_recover_ignores_unrelated_untracked_junk_exposed_by_missing_ignore_file ... FAILED
Capture worktree changes before recover the pre-undo state
Paths: unsaved worktree path(s): .venv/lib/junk.pyc
```

After the fix:

```text
$ heddle undo --output json
{"kind":"undo_requires_hard","error":"Undo would rewind repository state and worktree files","preserved":"repository state and worktree files were left unchanged",...}
plain_undo_exit=74
source_present_after_plain_refusal=yes
heddleignore_present_after_plain_refusal=yes

$ # with src/app.py changed but uncaptured
$ heddle undo --hard --output json
{"kind":"dirty_worktree","unsafe_condition":"unsaved worktree path(s): modified: src/app.py","preserved":"repository state and worktree files were left unchanged; no snapshot has been written for these paths",...}
dirty_hard_undo_exit=65
source_after_dirty_refusal=print("unsaved local edit")
heddleignore_present_after_dirty_refusal=yes

$ # after restoring the captured bytes
$ heddle undo --hard --output json
{"output_kind":"undo","status":"completed","recovery_marker":".undo-recovery",...}
hard_undo_exit=0
$ ls -la src
total 12
drwxr-xr-x 3 heddleco heddleco 4096 Jul 30 08:59 .
drwxr-xr-x 6 heddleco heddleco 4096 Jul 30 08:59 ..
drwxr-xr-x 2 heddleco heddleco 4096 Jul 30 08:59 __pycache__
heddleignore_survived_hard_undo=yes

$ # remove the preserved ignore file to recreate the former blocking state
$ heddle status --output json
{"summary":"1815 Heddle worktree path(s) are not captured in the current state",...}
$ heddle undo --recover --output json
{"output_kind":"undo_recover","status":"completed","message":"restored the state preserved by the most recent undo as worktree changes",...}
recover_exit=0
source_restored=yes
heddleignore_restored=yes
unrelated_venv_survived=yes
$ find .venv \( -type f -o -type l \) | wc -l
1814

$ heddle capture -m clear-recovery-blocker --output json
{"error":"symlink target escapes repository: .venv/bin/python -> python3","exit_code":74,...}
```

Automated verification:

```text
cargo test -p heddle-cli undo_apply -- --nocapture --test-threads=1
28 passed; 0 failed

cargo test -p heddle-repo --lib -- --test-threads=1
604 passed; 0 failed; 2 ignored

Focused CLI tests:
test_undo_capture_requires_hard_before_rewriting_worktree ... ok
test_undo_recover_round_trips_with_heddleignore_present ... ok
test_undo_recover_ignores_unrelated_untracked_junk_exposed_by_missing_ignore_file ... ok
test_undo_recover_refuses_without_recovery_state_or_with_overlapping_worktree_change ... ok
test_capture_escaping_symlink_error_names_path_and_target ... ok
test_undo_help_lists_undoable_and_unsupported ... ok

cargo clippy -p heddle-cli --all-targets -- -D warnings
Finished successfully
```

Formatting used `rustfmt +nightly --edition 2024` on the touched Rust files only. No ignore-resolution, discovery, config-resolution, or thread-lifecycle files were changed.

---
<a href="https://app.blacksmith.sh/HeddleCo/codesmith/heddle/pr/1160"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1787987445&installation_model_id=21489&pr_number=1160&repository=HeddleCo%2Fheddle&return_to=https%3A%2F%2Fgithub.com%2FHeddleCo%2Fheddle%2Fpull%2F1160&signature=24bd3195bdb35ff1045ae77409fe2752a433e2862d1166c40c28bb5cd1ff9395"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,11 @@ GitHub App, etc.) lives in the closed `HeddleCo/weft` and
 
 ### Changed
 
+- **Undo recovery no longer deadlocks on ignored local files.** Destructive
+  worktree rewinds now require `undo --hard`, preserve ignore-control files,
+  and leave unrelated local files available to `undo --recover`. Checkpoint-only
+  Git publication undo remains non-destructive and does not require `--hard`.
+
 - **Security boundaries hardened across local, hosted, and release paths.**
   Dynamic Bash completion no longer re-evaluates ref names; recursive clone
   mounts are confined beneath the clone root; hosted Git ref updates are
--- a/crates/cli-args/src/cli/cli_args/commands_args.rs
+++ b/crates/cli-args/src/cli/cli_args/commands_args.rs
@@ -577,8 +577,9 @@ pub struct RevertArgs {
 #[derive(Clone, Debug, clap::Args)]
 #[command(after_help = "\
 Examples:
-  heddle undo                # roll back the most recent operation
-  heddle undo -n 3           # roll back the last three operations
+  heddle undo --preview      # inspect the most recent operation
+  heddle undo --hard         # roll it back and rewind the worktree
+  heddle undo -n 3 --hard    # roll back the last three operations
   heddle undo --recover      # restore the state preserved by the last undo
   heddle undo --list         # preview undoable operations on this thread
   heddle undo --dry-run      # show what would change without applying
@@ -627,6 +628,12 @@ pub struct UndoArgs {
     #[arg(long, visible_alias = "dry-run")]
     pub preview: bool,
 
+    /// Permit undo to rewind worktree files to the selected operation's prior
+    /// state. Without this explicit opt-in, an undo that would rewrite the
+    /// worktree refuses before changing repository state or files.
+    #[arg(long, conflicts_with_all = ["list", "preview", "redo", "recover"])]
+    pub hard: bool,
+
     /// Re-apply operations that a prior `undo` rewound.
     #[arg(long, conflicts_with = "list")]
     pub redo: bool,
@@ -635,7 +642,7 @@ pub struct UndoArgs {
     /// worktree changes. HEAD and the current thread remain unchanged.
     #[arg(
         long,
-        conflicts_with_all = ["steps", "list", "preview", "redo", "allow_redact_undo"]
+        conflicts_with_all = ["steps", "list", "preview", "hard", "redo", "allow_redact_undo"]
     )]
     pub recover: bool,
 
--- a/crates/cli/src/cli/commands/undo.rs
+++ b/crates/cli/src/cli/commands/undo.rs
@@ -547,7 +547,32 @@ fn undo_may_rewrite_worktree(repo: &Repository, batches: &[OpBatch]) -> Result<b
             OpRecord::ThreadUpdate { name, .. } => active_thread
                 .as_ref()
                 .is_some_and(|thread| thread.as_str() == name.as_str()),
-            _ => false,
+            OpRecord::Snapshot {
+                prev_head: None, ..
+            }
+            | OpRecord::Goto {
+                prev_head: None, ..
+            }
+            | OpRecord::ThreadCreate { .. }
+            | OpRecord::ThreadDelete { .. }
+            | OpRecord::Fork { .. }
+            | OpRecord::Collapse { .. }
+            | OpRecord::MarkerCreate { .. }
+            | OpRecord::MarkerDelete { .. }
+            | OpRecord::Checkpoint { .. }
+            | OpRecord::TransactionAbort { .. }
+            | OpRecord::EphemeralThreadCollapse { .. }
+            | OpRecord::ConflictResolved { .. }
+            | OpRecord::TransactionCommit { .. }
+            | OpRecord::Redact { .. }
+            | OpRecord::Purge { .. }
+            | OpRecord::GitCheckpoint { .. }
+            | OpRecord::RemoteThreadUpdate { .. }
+            | OpRecord::RemoteThreadDelete { .. }
+            | OpRecord::UndoRecoveryUpdate { .. }
+            | OpRecord::StateVisibilitySet { .. }
+            | OpRecord::StateVisibilityPromote { .. }
+            | OpRecord::HeadUpdate { .. } => false,
         })
     }))
 }
--- a/crates/cli/src/cli/commands/undo.rs
+++ b/crates/cli/src/cli/commands/undo.rs
@@ -16,7 +16,7 @@ use heddle_core::{
     plan_undo_batches, summarize_batch, validate_undo_list_preview_modes,
 };
 use objects::store::ObjectStore;
-use oplog::OpBatch;
+use oplog::{OpBatch, OpRecord};
 use refs::UNDO_RECOVERY_HANDLE;
 use repo::{Repository, ThreadManager};
 use serde::Serialize;
@@ -93,6 +93,7 @@ pub(crate) fn undo_batches_quiet(repo: &Repository, batches: Vec<OpBatch>) -> Re
     ensure_redaction_undo_safe(repo, &batches, false)?;
     ensure_thread_worktree_undo_safe(repo, &batches)?;
     preflight_undo_execution(repo, &batches)?;
+    ensure_worktree_clean(repo, "undo")?;
     let recovery_state = repo.head()?;
     let scope = repo.op_scope();
     let generation = repo.oplog().head_id()?;
@@ -108,6 +109,7 @@ pub fn cmd_undo(
     list: bool,
     depth: usize,
     preview: bool,
+    hard: bool,
     allow_redact_undo: bool,
 ) -> Result<()> {
     let repo = cli.open_repo()?;
@@ -184,6 +186,24 @@ pub fn cmd_undo(
 
         return Ok(());
     }
+
+    if !hard && undo_may_rewrite_worktree(&repo, &apply_plan.batches)? {
+        return Err(anyhow!(RecoveryAdvice::safety_refusal(
+            "undo_requires_hard",
+            "Undo would rewind repository state and worktree files",
+            "Inspect the selected operation with `heddle undo --preview`, then rerun with `heddle undo --hard` to permit the worktree rewind.",
+            "the selected undo operation materializes an earlier saved tree",
+            "captured worktree files would be replaced by the selected operation's prior tree",
+            "repository state and worktree files were left unchanged",
+            "heddle undo --hard",
+            vec![
+                "heddle undo --preview".to_string(),
+                "heddle undo --hard".to_string(),
+            ],
+        )));
+    }
+
+    ensure_worktree_clean(&repo, "undo")?;
 
     // heddle#305: capture the pre-undo state into thread history BEFORE the
     // reset, so the worktree content the undone batch(es) absorbed is never
@@ -297,7 +317,6 @@ pub fn cmd_undo_recover(cli: &Cli) -> Result<()> {
         )));
     }
 
-    ensure_worktree_clean(&repo, "recover the pre-undo state")?;
     repo.restore_state_tree_to_worktree(&recovery_state)?;
 
     let recovered_repo = Repository::open(repo.root())?;
@@ -502,7 +521,6 @@ fn human_post_undo_trust_status(trust: &RepositoryVerificationState) -> String {
 
 fn preflight_undo_execution(repo: &Repository, batches: &[OpBatch]) -> Result<()> {
     ensure_no_active_operation(repo, "undo")?;
-    ensure_worktree_clean(repo, "undo")?;
     // Refuse before mutating anything when a state the batch needs to
     // restore is missing from the object store — typically because `gc
     // --prune` or a truncated oplog has reached past the live window.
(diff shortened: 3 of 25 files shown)


thethorne48:
Sending this back — the fix looks right but it breaks three existing undo tests, and the PR does not mention them.

```text
---- basics::test_cli_undo_changes_worktree ----
assertion failed: heddle(&["undo"], Some(temp.path())).is_ok()

---- basics::test_cli_undo_redo ----
assertion failed: heddle(&["undo"], Some(temp.path())).is_ok()

---- git_overlay_matrix::git_overlay_matrix_undo_reconciles_stale_mirror_after_push_pull ----
heddle ["undo"] failed: code=Some(74)
```

`undo` now returns an error where those tests expect success. That is very possibly **correct** — the whole point of the issue was that `undo` destroyed uncommitted work silently, so refusing in some cases is the intended new behaviour. But then those tests are asserting the old contract and need updating, with the change explained.

What I need before merging:

1. **Say which it is.** Either the tests encode the old destructive contract and should be updated — in which case update them and state what changed — or the new behaviour is refusing in a case it should not, and that is a bug in the fix.
2. If updating them, **do not just flip the assertion to `is_err()`.** Assert the new contract properly: that it refused, *and why*, and that the user's files survived. A test that only checks "it failed" would pass even if it failed for the wrong reason.
3. The `git_overlay_matrix` one is the one I would look at hardest. `code=Some(74)` in a push/pull reconciliation path suggests undo now refuses during a legitimate reconcile, which would be a real behavioural regression rather than a contract update.

The rest of the work is good and I do not want it redone: `recover_exit=0`, `source_restored=yes`, `heddleignore_restored=yes`, `unrelated_venv_survived=yes` is exactly the deadlock broken, and the symlink diagnostic naming both sides (`.venv/bin/python -> python3`) fixes the unactionable error. Both new tests demonstrated failing against the pre-fix binary, which is the evidence I asked for.

Note also that HeddleCo/heddle#1155 covers ignore handling separately, and you correctly did not touch ignore resolution — keep it that way.

forge-zephyr:
🤖 Starting. Worktree `workspace/issue-1160`, branch `task/1160-fix-undo-prevent-recovery-deadlock`. File scope: (inferred from epic; declare in PR).

forge-zephyr:
Rebased onto current `main` (`a5b5c75f`, including the merged discovery/sandbox fix) and pushed `5d13b9c6`.

Eight-way classification:

1. `basics::test_cli_undo_changes_worktree` — **(a), old destructive contract.** This is a `Snapshot` inverse that replaces `version.txt` v2 with v1. The test now proves plain undo exits 74 with `kind=undo_requires_hard`, names the materialization and replacement reason, leaves HEAD and v2 unchanged, then proves `undo --hard` restores v1.
2. `basics::test_cli_undo_redo` — **(a), old destructive contract.** The first inverse rewinds the captured tree. The test now proves the typed refusal, unchanged second-state HEAD and file bytes, then proves hard undo reaches the first state and redo returns to the second.
3. `git_overlay_matrix_undo_reconciles_stale_mirror_after_push_pull` — **(b), fix firing where it should not.** The selected batch is checkpoint-only. `GitCheckpoint` undo reconciles checkout Git HEAD/index and `.heddle/git` mirror refs; it intentionally does not materialize an earlier Heddle tree. The broad `GitCheckpoint => true` arm was the bug. I removed that arm. Plain undo succeeds again, both refs rewind, and the test now also proves `second.txt` survives.
4. `git_overlay_undo_recover_stays_coherent_without_git_on_path` — **(b).** Its first undo is also checkpoint-only. Requiring `--hard` blocked the intended no-Git publication rewind before the later recovery flow. Narrowing the detector restores plain undo and the existing test again proves recovery/worktree/Git coherence without `git` on PATH.
5. `git_replacement_matrix_commit_undo_leaves_capture_ready_to_recommit_without_git_on_path` — **(b).** This is explicitly a Git publication undo: rewind the visible Git ref/index but retain the captured Heddle work. Plain undo is correct. After narrowing, the test again proves `story.txt` survives and the capture remains ready to recommit without `git` on PATH.
6. `core_mutations_emit_post_verification_in_json` — **(b).** The undo exercised a checkpoint inverse, not a worktree materialization. The new guard incorrectly replaced the successful mutation record with a safety error. Narrowing restores the existing successful JSON and post-verification contract. No successful mutation output shape changed.
7. `folded_verb_flag_variants_emit_only_advertised_output_kinds` — **(a).** This fixture undoes a capture, so it really rewrites `a.txt`. It now proves the typed plain-undo refusal and that `a.txt` remains `two`, then uses `--hard` for the successful undo variants. Successful `undo`/`redo`/`undo_recover` output kinds remain exactly the advertised values.
8. `test_cli_pull_local_clean_active_checkout_materializes_before_publish` — **(a).** Undoing this active-thread `FastForward` calls the worktree materialization path, so explicit `--hard` is required. The test now proves plain undo refuses for the materialization reason while leaving the published thread ref and `remote` bytes unchanged, then proves hard undo restores the pre-pull ref and `base` bytes.

So category (b) was real: all four checkpoint-only cases were false positives from classifying every `GitCheckpoint` as a worktree rewrite. The implementation change is intentionally one line: checkpoint-only undo no longer requires `--hard`; `Snapshot`, `Goto`, active `ThreadUpdate`, and `FastForward` worktree rewinds still do.

The full run also exposed three additional old-contract tests outside the eight named failures (`visibility::undo_after_capture_with_nonpublic_default_reverts_snapshot_and_visibility_in_one_undo`, `comprehensive::integration::test_undo_redo_workflow`, and `multi_agent_worktrees::e2e::undo_is_scoped_to_the_current_thread`). Each now proves typed refusal plus unchanged HEAD/files (and unchanged folded visibility or sibling lane where applicable) before exercising `--hard`.

Machine contract: successful checkpoint undo still emits the existing `undo` payload; successful hard undo emits the same `undo` payload; no discriminator or mutation-response shape was changed. The refusal remains the typed exit-74 JSON error on stderr with empty stdout.

Verification:
- all eight named regressions plus the three additional old-contract tests pass
- CLI integration: 867 passed, 19 ignored
- core functionality: 93 passed
- comprehensive: 51 passed, 1 ignored
- multi-agent worktrees: 65 passed, 13 ignored
- complete affected non-CLI package and doctest set passed, including 633 repository tests
- `cargo clippy --locked -p heddle-cli --lib --bins --tests --examples -- -D warnings` passed
- targeted rustfmt check and `git diff --check` passed

No ignore-resolution logic was changed.

forge-zephyr:
Final follow-up after CI:

- Final head: `f9f8d4e27b63d461158bd4d0352c953b5fcf9383`
- Final Rust Tests run: https://github.com/HeddleCo/heddle/actions/runs/30573996335 — green, including Quality, CLI fault injection, Feature contracts, Postgres, FSKit/FUSE/ProjFS, and the aggregate Check & test gate.
- The eight-way classification and per-test contracts in https://github.com/HeddleCo/heddle/pull/1160#issuecomment-5134824540 remain unchanged.

The first CI pass exposed one additional old-contract fixture after the Rust tests had passed: both default cold-flow scripts captured `native-authority-proof.txt` and then expected plain `undo` to delete it. They now encode the full safety contract on both the human and agent flows: plain undo exits 74, JSON stdout is empty, the envelope names `undo_requires_hard`, the exact unsafe condition / prospective change / preservation guarantee / primary command are asserted, and `native-authority-proof.txt` is verified byte-for-byte unchanged before `undo --hard` exercises the intended rewind.

The second CI pass exposed the repository's OpRecord exhaustiveness gate. `undo_may_rewrite_worktree` now lists every non-rewriting operation explicitly instead of using `_ => false`. That keeps `GitCheckpoint` explicitly outside the worktree-rewind policy and ensures any future OpRecord variant must deliberately choose whether it can materialize files.

Final local verification:

- `cargo test --locked -p heddle-cli --test cli_integration`: 867 passed, 19 ignored
- targeted CLI/devtools clippy with `-D warnings`: passed
- OpRecord production-tree exhaustiveness test: passed
- `scripts/check-default-cli-contracts.sh`: passed end-to-end for human and agent cold flows plus schema verification

No ignore-resolution code was changed, and nothing was merged.

forge-zephyr:
♻️ Released by orchestrator. Reason: manual release. Worktrees preserved for inspection: `workspace/issue-1160`.

Issue #892, opened 2026-07-02
WEFT? no — HEDDLE-DR-14 [P2] OpRecord thread-name projection re-derived (divergent) in weft


thethorne48:
**Crate:** schema — **Lens:** betterpath

- **file:** `crates/schema/src/op_record/types.rs:426`
- **evidence:** schema owns `isolation_keys_for_record(..)`; weft `operation_log_query.rs:179 fn thread_for(op)` re-implements a *different* projection — it returns `MarkerCreate/MarkerDelete name` as a thread, which `isolation_keys_for_record` deliberately does not.
- **what:** "Which thread does this op concern" is a merge/query-time semantic that belongs on the intent primitive but lives re-derived in the consumer's presentation layer — and the two already disagree on markers.
- **fix:** Add `OpRecord::thread_name(&self) -> Option<&str>` (exhaustive, beside `verb`/`description`); have weft's `thread_for` call it.
- **vision:** #4/#5 — provenance loss across the wire boundary; the #852/#856 pattern.

Source: 2026-07-01 heddle deep review (deep-review-2026-07-01/heddle-deep-review.md), lens: betterpath.

PR #1163, opened 2026-07-30
feat: add client-authorized provider pull path

forge-zephyr:
## Summary

- add the client-authorized two-phase provider negotiation to the existing signed pull stream: the opening binds version, actual client Iroh endpoint, and a one-use nonce; the second signature consents to one exact grant-batch digest with the same cnf key
- add one pooled Iroh connection per cryptographic provider endpoint over a bounded WSS custom transport, reuse it across extents, and preserve custom-transport backpressure
- download only opaque logical extents, resume from server-selected offsets with retained-prefix rehashing, and send the final digest only after the complete extent verifies
- assemble a virtual LMPK from storage-coordinate-free output extents, compute the 32-byte BLAKE3 trailer locally, validate the container and index, and install nothing until the logical pull completes
- silently retain the existing multi-Want Weft path when provider negotiation is absent, declined, unavailable, interrupted, malformed, or digest-mismatched; callers still receive one PullComplete

Merge order: 1. HeddleCo/api#62, 2. this Heddle client/fallback PR, 3. the Weft server/Worker integration in HeddleCo/weft#892. This client safely half-closes into the current ordinary pack stream when Weft does not challenge; Weft must only challenge openings that advertise the v1 capability context.

The Iroh fork is needed client-side. This PR pins HeddleCo/iroh#1 at `6d50cebc0a18a8b2d2cc9303d938bd5fa39fffd6` and enables `unstable-custom-transports` plus `unstable-custom-transport-backpressure`; it does not enable the Worker-only custom runtime hook. The authoritative document names base commit `e8415b1ff21f104504f76feac030860c8e7beae4`, but that commit does not contain the requested backpressure feature, so the document's SHA and feature statement are internally inconsistent.

Weft/Worker contract required by this client:

- decode provider capability context only after verifying StreamOpeningProof; require v1, a 16-byte one-use nonce, and equality between the claimed client endpoint and the transport-observed peer, while binding grants only to the observed endpoint
- independently authorize the closed multi-Want object plan with the owner-rooted capability; the cnf signature is consent and possession proof, never repository authorization
- construct one immutable private grant batch and domain-separated BLAKE3 digest covering subject, repository, observed endpoint, nonce, provider endpoint, private R2 key/ETag/source range/digest, output layout/object identity, and expiry; Worker must use the same canonical encoder
- challenge before emitting ordinary transfer frames, with exact provider-only object/extent/byte totals and expiry but no R2 coordinates; verify an accepted response against the same cnf key, nonce, and digest
- forward the exact private batch, client signature, and owner chain to the broker/Worker; on success send a manifest echoing the digest/nonce with a complete 16-byte LMPK v3 header, output-only extents and object offsets, digests, endpoint, direct WSS URL, opaque ticket, and expiry
- accept `provider/1` Iroh streams containing a ticket-only read request; return server-selected resumeOffset, a fresh nonzero attemptGeneration, and exact remaining length; reject stale generations
- stream the declared raw body, treat ordinary checkpoints as resumable progress only, and commit completion only after the exact final 32-byte digest; failure or mismatch must leave the completed/committed offset unchanged
- after a decline, malformed/missing response, timeout, provider FALLBACK result, or any install failure, install no Worker grant and continue the same authorized request through the existing multi-Want Pack/sidecar/Complete path
- after provider COMPLETE, require the exact client-computed LMPK trailer digest before treating the provider transfer as complete; provider and ordinary packs may both be emitted for one logical pull
- do not wait for request EOF before sending a challenge or manifest; the client half-closes after decline/result or when ordinary transfer begins

Deferred until the parallel Weft half is available: a live cross-repository Weft/broker/Cloudflare Worker end-to-end run. The provider framing, direct Iroh transfer, resume/digest behavior, virtual-pack assembly, and ordinary fallback decision are covered locally; production grant installation and live WSS transport remain the integration gate for HeddleCo/weft#892.

## Closes

Refs HeddleCo/weft#892

## Test evidence

- `CARGO_HOME=/home/scratch/heddle-provider/cargo-home TMPDIR=/home/scratch cargo build --locked --workspace --tests --bin heddle --bin heddle-fuse-worker`
- `CARGO_HOME=/home/scratch/heddle-provider/cargo-home TMPDIR=/home/scratch cargo clippy --locked --workspace --lib --bins --tests --examples -- -D warnings`
- `CARGO_HOME=/home/scratch/heddle-provider/cargo-home TMPDIR=/home/scratch cargo test --locked --workspace`
- `cargo test --locked -p heddle-client provider_ -- --nocapture`: 6 passed; digest mismatch rejected with committed offset 0 and no final ack, captured proof/different-plan replay rejected, interrupted transfer resumed at byte 24 after rehashing all 24 retained bytes, provider unavailability selected FALLBACK, and one provider connection was reused
- `cargo test --locked -p heddle-wire provider_pack -- --nocapture`: 3 passed; provider and ordinary paths produced identical 142-byte packs and 98-byte indexes with trailer `866c1fc447fa044ae18ce3f6c92c1bce04be13fcc7abeecd03adfebdbb94e6ef`

---
<a href="https://app.blacksmith.sh/HeddleCo/codesmith/heddle/pr/1163"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1788036431&installation_model_id=21489&pr_number=1163&repository=HeddleCo%2Fheddle&return_to=https%3A%2F%2Fgithub.com%2FHeddleCo%2Fheddle%2Fpull%2F1163&signature=8b592660ce7a276e38c955164e1b8e33178fd3db7db4bb5001bab7cf0be83405"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2139,7 +2139,7 @@ dependencies = [
 [[package]]
 name = "heddle-api"
 version = "0.2.0"
-source = "git+https://github.com/HeddleCo/api?rev=4f032f4e25e8a2e1a211cd8bfe530dc867d5ca7e#4f032f4e25e8a2e1a211cd8bfe530dc867d5ca7e"
+source = "git+https://github.com/HeddleCo/api?rev=88e3bce2152a3aa8c5dea8908b16efd694c3ef76#88e3bce2152a3aa8c5dea8908b16efd694c3ef76"
 dependencies = [
  "bytes",
  "hex",
@@ -2165,7 +2165,7 @@ dependencies = [
  "clap_complete",
  "criterion",
  "futures",
- "heddle-api 0.2.0 (git+https://github.com/HeddleCo/api?rev=4f032f4e25e8a2e1a211cd8bfe530dc867d5ca7e)",
+ "heddle-api 0.2.0 (git+https://github.com/HeddleCo/api?rev=88e3bce2152a3aa8c5dea8908b16efd694c3ef76)",
  "heddle-cli-args",
  "heddle-cli-contract",
  "heddle-cli-render",
@@ -2302,10 +2302,11 @@ dependencies = [
  "anyhow",
  "base64 0.22.1",
  "biscuit-auth",
+ "blake3",
  "bytes",
  "chrono",
  "futures",
- "heddle-api 0.2.0 (git+https://github.com/HeddleCo/api?rev=4f032f4e25e8a2e1a211cd8bfe530dc867d5ca7e)",
+ "heddle-api 0.2.0 (git+https://github.com/HeddleCo/api?rev=88e3bce2152a3aa8c5dea8908b16efd694c3ef76)",
  "heddle-cli-shared",
  "heddle-crypto",
  "heddle-objects",
@@ -2314,6 +2315,9 @@ dependencies = [
  "heddle-wire",
  "hex",
  "iroh",
+ "iroh-base",
+ "n0-watcher",
+ "noq-udp",
  "prost 0.14.1",
  "prost-reflect",
  "prost-types 0.14.1",
@@ -2385,7 +2389,7 @@ version = "0.10.4"
 dependencies = [
  "chrono",
  "futures",
- "heddle-api 0.2.0 (git+https://github.com/HeddleCo/api?rev=4f032f4e25e8a2e1a211cd8bfe530dc867d5ca7e)",
+ "heddle-api 0.2.0 (git+https://github.com/HeddleCo/api?rev=88e3bce2152a3aa8c5dea8908b16efd694c3ef76)",
  "heddle-crypto",
  "heddle-objects",
  "heddle-oplog",
@@ -2736,6 +2740,7 @@ dependencies = [
 name = "heddle-wire"
 version = "0.10.4"
 dependencies = [
+ "blake3",
  "chrono",
  "heddle-objects",
  "heddle-repo",
@@ -3231,8 +3236,8 @@ dependencies = [
 
 [[package]]
 name = "iroh"
-version = "1.0.2"
-source = "git+https://github.com/n0-computer/iroh.git?rev=cc876c75e53171dfcaf2e5fa0b81a417904694ba#cc876c75e53171dfcaf2e5fa0b81a417904694ba"
+version = "1.0.3"
+source = "git+https://github.com/HeddleCo/iroh.git?rev=6d50cebc0a18a8b2d2cc9303d938bd5fa39fffd6#6d50cebc0a18a8b2d2cc9303d938bd5fa39fffd6"
 dependencies = [
  "backon",
  "blake3",
@@ -3280,8 +3285,8 @@ dependencies = [
 
 [[package]]
 name = "iroh-base"
-version = "1.0.2"
-source = "git+https://github.com/n0-computer/iroh.git?rev=cc876c75e53171dfcaf2e5fa0b81a417904694ba#cc876c75e53171dfcaf2e5fa0b81a417904694ba"
+version = "1.0.3"
+source = "git+https://github.com/HeddleCo/iroh.git?rev=6d50cebc0a18a8b2d2cc9303d938bd5fa39fffd6#6d50cebc0a18a8b2d2cc9303d938bd5fa39fffd6"
 dependencies = [
  "curve25519-dalek 5.0.0",
  "data-encoding",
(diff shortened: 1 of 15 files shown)


thethorne48:
Weft #892 integration note (exact head `a7268c5e937f05ebd06b0f5f73328aa76bff86f0`): the v1 opening/challenge/consent/manifest/result contract now compiles through the Weft pull state machine, including ordinary fallback and mixed Worker+Weft objects.

One Heddle follow-up is required for the production throughput/memory target, though it is not blocking protocol-correct MVP testing: `HostedClient::download_provider_pull` currently awaits each extent sequentially and retains every extent body in `Vec<Vec<u8>>` before `assemble_provider_pack`. That means the current client does not yet realize multi-DO fan-out and its peak memory scales with the complete provider pack.

The production shape should use bounded concurrent extent downloads (group/reuse one connection per provider endpoint, but allow independent provider endpoints to progress in parallel), write verified extents directly into a positional spool/pack sink, then validate/install only after complete coverage and trailer verification. Resume generation/digest invariants must remain per extent. I am keeping Weft correctness independent of this so we can finish the live cross-repo MVP first, but this is a gate for the >R2-speed fan-out claim.

thethorne48:
Consolidated Weft #892 integration review for Heddle #1163:

MVP correctness status

- There is no remaining Heddle API or protocol-correctness blocker for the single-provider MVP. A normal pull internally handles challenge, client consent signature, provider manifest, provider result, and ordinary Weft fallback without a second authorization request or a caller-visible redirect.
- Keep the Worker request contract to provider endpoint plus opaque ticket. Do not add clientKey or other caller-asserted identity parameters; the provider must bind the installed grant to the authenticated Iroh peer.
- The client signature must continue to cover the exact canonical provider plan, verified stream/repository binding, endpoint, nonce, expiry, and batch digest. A retry needs a fresh challenge/nonce; a signed plan must never be editable by the caller.

Production throughput and memory blockers

1. Provider extents are currently awaited sequentially. This prevents distinct DO endpoint IDs from progressing in parallel, so a four-way manifest still behaves like one lane.
2. Every completed extent is retained in Vec<Vec<u8>>, followed by assembly into a second full-pack allocation. Peak RAM is therefore roughly extent bodies plus the assembled pack, and the hot path copies the entire payload again.
3. Sequential scheduling interacts badly with the roughly 60-second grant lifetime and large manifests: later grants can expire before Heddle attempts them even when every provider is healthy.
4. Connection setup should be amortized by grouping grants by provider endpoint ID and reusing one connection per endpoint. Distinct endpoint IDs should run concurrently; multiple addresses for one endpoint remain alternative paths, not separate providers.
5. Backpressure should be end-to-end. Bound both active provider connections and in-flight bytes so concurrency does not turn into an unbounded memory queue.

Recommended implementation shape

- Build a bounded scheduler over provider groups, with configurable global and per-provider concurrency.
- Pre-size a positional tempfile/spool for the exact virtual pack length. Stream each verified extent directly to its assigned offset instead of collecting bodies.
- Reject overlap, holes, length mismatches, unexpected object/range bindings, duplicate completion, or bytes beyond the planned pack before installation.
- Validate the final pack/index/trailer only after complete coverage, then install atomically. Partial provider output must never become a valid partial pack.
- On expiry, stall, or provider error, cancel outstanding provider work and use the already-established Weft fallback path. Do not combine unverified partial provider bytes with fallback bytes.
- Start eligible provider groups promptly enough that grant expiry measures authorization lifetime rather than queue residence. If a grant expires before start, fall back; do not silently extend or replay it.

Tests that should gate the production claim

- Two or more distinct provider endpoint IDs complete out of order while reconstructing a byte-identical native pack.
- Multiple grants for one endpoint reuse one connection.
- Peak buffered memory stays bounded as pack size grows.
- Slow/stalled lane does not block healthy lanes until the configured fallback threshold.
- Expired, overlapping, gapped, truncated, oversized, and digest-mismatched extents fail closed and exercise Weft fallback.
- Cancellation drops in-flight provider streams and temporary spool state.
- The public Heddle pull API observes one logical pull and no caller-managed second request.

The Worker v3 verifier/route and direct provider serving implementation remain the live E2E blocker, but that is not a Heddle contract change. The Heddle work above is the gate for claiming bounded-memory multi-DO fan-out or greater-than-single-R2-lane throughput.

thethorne48:
Weft hardening follow-up for the Worker v3 boundary:

- Weft now enforces fail-fast per-process negotiation concurrency plus per-pull record and byte ceilings. The broker Durable Object must still enforce the authoritative cross-instance weighted quota, keyed by tenant/principal and charged by admitted records plus bytes. Without that broker-side check, horizontal Weft replicas multiply the local allowance and one tenant can occupy a process-wide budget.
- Before Weft attaches the provider path, it now performs an authenticated GET /healthz against the configured broker origin. The Worker must return bounded JSON with status=ok, service=weft-iroh-grant-broker, and contractVersion=3. Additive metadata is allowed. Redirects are never followed.
- The grant-install endpoint must continue to apply the same broker-side quota atomically with installation; a readiness response alone is not authorization or admission.

These are Worker/broker deployment requirements, not changes to Heddle’s one-logical-pull client abstraction. The ordinary Weft fallback remains available when readiness, admission, or grant installation fails.

PR #1166, opened 2026-07-31
Provider pull: bounded concurrent fan-out and streaming assembly

### LINKED ISSUE #1165: Provider pull: bounded concurrent fan-out and streaming assembly (throughput/memory gate)
Follow-up to #1163, from the consolidated Weft #892 integration review posted at
https://github.com/HeddleCo/heddle/pull/1163#issuecomment-5137191308.

## Not in scope: protocol correctness

Per that review, **there is no remaining Heddle API or protocol-correctness blocker for the
single-provider MVP.** A normal pull already handles challenge, client consent signature, provider
manifest, provider result, and Weft fallback internally, with no second authorization request and no
caller-visible redirect. This issue does not reopen any of that.

Three contract properties must survive every change here:

- The Worker request contract stays **provider endpoint + opaque ticket**. Do not add `clientKey` or
  any other caller-asserted identity parameter; the provider binds the installed grant to the
  authenticated Iroh peer.
- The client signature must continue to cover the exact canonical provider plan, verified
  stream/repository binding, endpoint, nonce, expiry, and batch digest.
- A retry needs a **fresh challenge/nonce**. A signed plan must never be editable by the caller.

## The problem

The current client is correct but serialised, and it buffers everything.

1. **Provider extents are awaited sequentially.** Distinct DO endpoint IDs cannot progress in
   parallel, so a four-way manifest behaves like a single lane.
2. **Every completed extent is retained in `Vec<Vec<u8>>`**, then assembled into a second full-pack
   allocation. Peak RAM is roughly *extent bodies + the assembled pack*, and the hot path copies the
   whole payload again.
3. **Sequential scheduling interacts badly with the ~60 second grant lifetime.** On a large manifest,
   later grants can expire before Heddle ever attempts them, even when every provider is healthy.
   Grant expiry should measure authorization lifetime, not queue residence.
4. **Connection setup is not amortised.** Grants should be grouped by provider endpoint ID with one
   reused connection per endpoint. Multiple addresses for one endpoint are alternative paths, not
   separate providers.
5. **Backpressure is not end-to-end.** Without bounding both active connections and in-flight bytes,
   concurrency just converts into an unbounded memory queue.

## Implementation shape

- A **bounded scheduler over provider groups**, with configurable global and per-provider concurrency.
- **Pre-size a positional tempfile/spool** for the exact virtual pack length and stream each verified
  extent directly to its assigned offset. Do not collect bodies.
- **Reject before installation**: overlap, holes, length mismatches, unexpected object/range bindings,
  duplicate completion, or any bytes beyond the planned pack.
- **Validate pack/index/trailer only after complete coverage**, then install atomically. Partial
  provider output must never become a valid partial pack.
- **On expiry, stall, or provider error**: cancel outstanding provider work and use the existing Weft
  fallback. Never combine unverified partial provider bytes with fallback bytes.
- If a grant expires before start, **fall back** — do not silently extend or replay it.

## Gating tests

These decide whether the production claim can be made:

1. Two or more distinct provider endpoint IDs complete **out of order** and still reconstruct a
   byte-identical native pack.
2. Multiple grants for one endpoint **reuse one connection**.
3. **Peak buffered memory stays bounded as pack size grows** — this is the whole point; a test that
   only checks correctness does not gate this issue.
4. A slow or stalled lane **does not block healthy lanes** until the configured fallback threshold.
5. Expired, overlapping, gapped, truncated, oversized, and digest-mismatched extents **fail closed**
   and exercise Weft fallback.
6. Cancellation drops in-flight provider streams **and** temporary spool state.
7. The public Heddle pull API observes **one logical pull**, with no caller-managed second request.

## Relationship to other work

- Builds on #1163; branch from that PR's head rather than `main`, since it is not yet merged
  (#1163 is itself gated on `HeddleCo/api#62`, which is red pending a `PROJECT_PAT` secret).
- The Worker v3 verifier/route and direct provider serving remain the live E2E blocker, but that is a
  **Weft/Worker** concern and not a Heddle contract change. The work in this issue is what gates any
  claim of bounded-memory multi-DO fan-out or throughput beyond a single R2 lane.

forge-zephyr:
## Summary

- Stacks on #1163 and keeps its provider endpoint + opaque ticket contract, canonical plan signature, and fresh challenge/nonce boundary unchanged.
- Schedule provider endpoint groups concurrently with one cached connection per endpoint, global/per-endpoint stream limits, grant-expiry checks at actual start, progress-stall fallback, and a shared in-flight body-byte budget.
- Pre-size an exact-length positional spool, stream verified extent chunks directly to their assigned ranges, validate complete coverage/index/trailer, and install only the completed pack through the atomic streaming store path.

Defaults are 4 global streams, 2 streams per endpoint, an 8 MiB provider-body budget, and a 15 second progress-stall threshold. Four starts a typical four-lane plan promptly within the roughly 60 second grant lifetime; two gives useful multiplexing while amortising one endpoint connection; 8 MiB leaves headroom for the default 1 MiB read chunks without tying memory to pack size; 15 seconds leaves time to select the existing Weft fallback before grant expiry.

## Closes

Closes #1165

## Test evidence

1. Out-of-order fan-out: `distinct_endpoints_complete_out_of_order_into_a_byte_identical_pack` completed endpoint B before endpoint A and reconstructed the exact 262,270-byte native pack. The wire positional-spool byte-identity test also passed.
2. Connection reuse: `one_endpoint_reuses_one_connection_for_multiple_grants` observed 3 grants and exactly 1 connection; the real hosted connection cache test also observed 1 connection by cryptographic endpoint ID.
3. Bounded peak memory: `peak_provider_body_memory_is_constant_as_pack_size_grows` measured the production live-body-buffer high-water mark. A 1,048,784-byte pack peaked at 524,288 live provider-body bytes; a 16,777,428-byte pack peaked at the same 524,288 bytes. After each transport read, production records the actual `Bytes::len()` and keeps it live through digesting, positional spool write, and checkpoint; drop subtracts that exact payload length. The test backend allocates matching `Bytes` chunks and yields after allocation so concurrent buffers coexist. This measures live application provider-payload bytes, excluding fixture storage, allocator metadata, process RSS, and filesystem cache. The semaphore reservation is separately acquired before each read, so it remains the hard backpressure bound. Negative control: temporarily replacing the configured production budget with an unbounded budget made this actual-buffer test fail; restoring the bound made it pass. Compression is disabled in the fixture so the sizes differ by more than an order of magnitude on the wire.
4. Slow lane isolation: `stalled_lane_does_not_block_healthy_lane_before_fallback_threshold` completed the healthy lane and selected fallback after 101 ms with a configured 100 ms threshold.
5. Fail closed: `invalid_provider_extents_all_select_existing_weft_fallback` covers expired, overlapping, gapped, truncated, oversized, and digest-mismatched cases. Wire tests separately reject range overruns, duplicate completion, incomplete coverage, invalid object bindings, and digest mismatch before an installable pack exists.
6. Cancellation: `cancellation_drops_streams_and_partial_spool_state` observed 2 in-flight operations, all 3 streams dropped, and 0 remaining spool entries.
7. One logical pull: `fallback_stays_on_the_same_logical_pull_and_carries_no_success_digest` observed one logical pull, the same nonce and batch binding, unchanged caller protocol, and no provider success digest on fallback.

Focused evidence after final formatting:

- `cargo test --locked -p heddle-client provider_ -- --nocapture` — 12 passed.
- `cargo test --locked -p heddle-wire provider_pack::tests -- --nocapture` — 6 passed.

Exact Rust workflow invocations completed on the final code:

- `cargo build --locked --workspace --tests --bin heddle --bin heddle-fuse-worker` — passed, `dev` profile.
- `cargo clippy --locked --workspace --lib --bins --tests --examples -- -D warnings` — passed.
- `cargo test --locked --workspace` — passed, including doctests.
- `cargo check --locked -p heddle-repo --no-default-features --features git-overlay,zstd` — passed.
- `cargo check --locked -p heddle-repo --no-default-features --features native,zstd` — passed.
- `cargo check --locked -p heddle-cli --no-default-features --features git-overlay,client,semantic,zstd` — passed with existing conditional-compilation warnings.
- `cargo check --locked -p heddle-cli --no-default-features --features native,semantic,zstd` — passed with existing conditional-compilation warnings.

All commands used `TMPDIR=/home/scratch`. The unrelated red `HeddleCo/api#62` repository-secret gate was not changed.

### DIFF
--- a/crates/cli-shared/src/client_config.rs
+++ b/crates/cli-shared/src/client_config.rs
@@ -53,6 +53,14 @@ pub struct ClientConfig {
     pub pack_transfer: bool,
     /// Enable partial fetch negotiation.
     pub partial_fetch: bool,
+    /// Maximum provider extent streams active across all provider endpoints.
+    pub provider_global_concurrency: usize,
+    /// Maximum provider extent streams multiplexed over one endpoint connection.
+    pub provider_per_endpoint_concurrency: usize,
+    /// Maximum provider body bytes retained between transport reads and spool writes.
+    pub provider_max_inflight_bytes: usize,
+    /// Maximum seconds without provider transport progress before falling back.
+    pub provider_stall_timeout_secs: u64,
 }
 
 impl ClientConfig {
@@ -78,6 +86,10 @@ impl ClientConfig {
             resumable_transfer: true,
             pack_transfer: true,
             partial_fetch: true,
+            provider_global_concurrency: 4,
+            provider_per_endpoint_concurrency: 2,
+            provider_max_inflight_bytes: 8 * 1024 * 1024,
+            provider_stall_timeout_secs: 15,
         }
     }
 
@@ -175,6 +187,30 @@ impl ClientConfig {
         self
     }
 
+    /// Set the maximum number of provider extent streams active globally.
+    pub fn with_provider_global_concurrency(mut self, concurrency: usize) -> Self {
+        self.provider_global_concurrency = concurrency.max(1);
+        self
+    }
+
+    /// Set the maximum provider extent streams multiplexed over one endpoint.
+    pub fn with_provider_per_endpoint_concurrency(mut self, concurrency: usize) -> Self {
+        self.provider_per_endpoint_concurrency = concurrency.max(1);
+        self
+    }
+
+    /// Set the provider body-byte budget shared by all active extent streams.
+    pub fn with_provider_max_inflight_bytes(mut self, bytes: usize) -> Self {
+        self.provider_max_inflight_bytes = bytes.max(1);
+        self
+    }
+
+    /// Set the provider transport progress threshold before ordinary Weft fallback.
+    pub fn with_provider_stall_timeout(mut self, secs: u64) -> Self {
+        self.provider_stall_timeout_secs = secs.max(1);
+        self
+    }
+
     /// Explicitly allow cleartext to non-loopback addresses.
     pub fn with_allow_insecure(mut self, allow: bool) -> Self {
         self.allow_insecure = allow;
@@ -243,5 +279,24 @@ mod tests {
         assert!(is_loopback_ip(IpAddr::V4(Ipv4Addr::new(127, 0, 0, 1))));
         assert!(is_loopback_ip(IpAddr::V6(Ipv6Addr::LOCALHOST)));
         assert!(!is_loopback_ip(IpAddr::V4(Ipv4Addr::new(10, 0, 0, 1))));
+    }
+
+    #[test]
+    fn provider_scheduler_defaults_and_builders_are_explicit() {
+        let defaults = ClientConfig::default();
+        assert_eq!(defaults.provider_global_concurrency, 4);
+        assert_eq!(defaults.provider_per_endpoint_concurrency, 2);
+        assert_eq!(defaults.provider_max_inflight_bytes, 8 * 1024 * 1024);
+        assert_eq!(defaults.provider_stall_timeout_secs, 15);
+
+        let configured = ClientConfig::default()
+            .with_provider_global_concurrency(8)
+            .with_provider_per_endpoint_concurrency(3)
+            .with_provider_max_inflight_bytes(4 * 1024 * 1024)
+            .with_provider_stall_timeout(20);
+        assert_eq!(configured.provider_global_concurrency, 8);
+        assert_eq!(configured.provider_per_endpoint_concurrency, 3);
+        assert_eq!(configured.provider_max_inflight_bytes, 4 * 1024 * 1024);
+        assert_eq!(configured.provider_stall_timeout_secs, 20);
     }
 }
--- a/crates/cli-shared/src/config/user_config.rs
+++ b/crates/cli-shared/src/config/user_config.rs
@@ -165,6 +165,14 @@ pub struct UserRemoteConfig {
     pub iroh_descriptor_key_id: Option<String>,
     #[serde(default)]
     pub iroh_descriptor_public_key_path: Option<PathBuf>,
+    #[serde(default)]
+    pub provider_global_concurrency: Option<usize>,
+    #[serde(default)]
+    pub provider_per_endpoint_concurrency: Option<usize>,
+    #[serde(default)]
+    pub provider_max_inflight_bytes: Option<usize>,
+    #[serde(default)]
+    pub provider_stall_timeout_secs: Option<u64>,
     /// Allow cleartext connections to non-loopback hosts without TLS.
     /// Prefer enabling TLS; this is an explicit opt-in for lab/VPN testing.
     #[serde(default)]
@@ -486,6 +494,18 @@ impl UserConfig {
                 "both descriptor trust fields are required".to_string(),
             ));
         }
+        if let Some(value) = self.remote.provider_global_concurrency {
+            config = config.with_provider_global_concurrency(value);
+        }
+        if let Some(value) = self.remote.provider_per_endpoint_concurrency {
+            config = config.with_provider_per_endpoint_concurrency(value);
+        }
+        if let Some(value) = self.remote.provider_max_inflight_bytes {
+            config = config.with_provider_max_inflight_bytes(value);
+        }
+        if let Some(value) = self.remote.provider_stall_timeout_secs {
+            config = config.with_provider_stall_timeout(value);
+        }
 
         if env_bool("HEDDLE_REMOTE_TLS")? {
             config = config.with_tls(false);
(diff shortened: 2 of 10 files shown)


forge-zephyr:
## Review: the seven gates are met, and the memory claim is properly falsified

Checked against the PR, the diff, and CI rather than the summary. **CI rollup is fully green** — the
only such PR across the org today.

The gate I cared about is item 3, and it was done right:

- 1,048,784-byte pack → peak **524,288** live provider-body bytes
- 16,777,428-byte pack → peak **524,288**

16× the pack, identical peak. Crucially there is a **negative control**: replacing the configured
budget with an unbounded one made the test fail, and restoring it made it pass. That is the difference
between a measurement and an assertion, and it is what I asked for.

The exclusions are stated honestly too — application live-payload bytes, not RSS, allocator metadata,
or filesystem cache. That is the right thing to measure, because it is precisely what the bound
governs.

Item 4's isolation result (healthy lane completes, fallback selected at 101 ms against a 100 ms
threshold) and item 6's cancellation result (2 in-flight, 3 streams dropped, 0 spool entries) are
likewise concrete rather than narrated.

## Two observations, neither blocking

### 1. Bounded against pack *size*, not extent *count*

`peak_provider_body_memory_is_constant_as_pack_size_grows` holds the manifest at **four** extents
across four endpoints in both measurements and scales only the payload per extent. So it proves peak
body buffering is independent of pack size — which is exactly the defect this PR set out to fix, since
the old code retained every body in `Vec<Vec<u8>>`.

What it does not exercise is the **other axis**: a manifest with hundreds of extents at constant total
size. Per-extent state — manifest entries, digests, checkpoint records, positional-spool bookkeeping —
scales with count, not bytes, and none of it is covered by `peak_inflight_bytes`.

That is a plausible future problem rather than a present one, and the review's stated target was
multi-DO fan-out where extent counts are modest. Worth a follow-up test that scales extent count with
total size held constant, so the second axis is measured before someone assumes it was.

### 2. The 101 ms / 100 ms margin will flake

Item 4 asserts fallback selection at 101 ms against a configured 100 ms threshold. A one-millisecond
margin on a shared CI runner under load will eventually fail for reasons unrelated to the code, and
that failure will look like a real regression in exactly the lane whose whole purpose is timing.

Prefer virtual time, or a margin wide enough that scheduling jitter cannot cross it.

## Merge chain unchanged

`HeddleCo/api#62` → `#1163` → **this**. api#62 is still red on the missing `PROJECT_PAT` repository
secret, which is operator work and unrelated to this code.

No settled protocol property changed, and the Worker v3 E2E and api#62 gate were correctly left alone.

thethorne48:
## Weft/Worker integration requirements

The fan-out and positional-spool shape aligns with the provider design, but the following changes are required for the Heddle side to match the server contract Weft and the Worker will enforce.

### 1. Treat signed grant expiry as the absolute plan deadline

Weft waits for `ProviderPullResponse` only until the challenge/grant expiry. The Worker will reject new/resumed reads after expiry and terminate an in-flight read when that absolute deadline arrives. Checking expiry only before an extent starts is therefore insufficient: a continuously-progressing transfer must not run past the signed deadline.

- Bound connect + all extent reads + spool finalization/install by the minimum signed source/plan expiry.
- Preserve enough control-plane margin to send `FALLBACK` before Weft stops waiting.
- On deadline, cancel all lanes, remove partial spool state, and use the existing same-pull Weft fallback. Do not perform another authorization request.
- Do not report `COMPLETE` after the deadline.

Regression: a provider that continuously sends sub-stall chunks beyond expiry must be cancelled and select same-pull fallback before expiry; no success digest may be sent.

### 2. Resume through a replacement carrier connection

The current group connects once and all resume attempts call `open_bi()` on the same cloned `iroh::Connection`. That handles a reset stream on a live connection, but not WebSocket/QUIC carrier loss or DO migration.

- On a retryable carrier/connection failure, reacquire the connection through `HostedConnection::provider_connection(source)` so the closed cache entry is replaced.
- Keep the same cryptographic endpoint ID and opaque grant; resume from the Worker-returned committed offset.
- A malformed resume response remains fatal and falls back.

Regression: transfer a prefix, close the entire carrier, reconnect to the same endpoint ID, receive the prior committed offset/generation, and complete byte-identically without a second Weft authorization.

### 3. Install before reporting provider `COMPLETE`

`COMPLETE` should mean the provider pack has been validated and accepted into the local object store, not merely written to a temporary spool.

- Run `install_pack_streaming` before sending `ProviderPullResultStatus::Complete`.
- If validation/install fails, send same-pull `FALLBACK` while Weft can still provide the ordinary pack.
- CAS objects installed before a later Weft stream failure are harmless and reusable on retry.

Regression: force `install_pack_streaming` to fail and assert Heddle sends `FALLBACK`, never `COMPLETE`.

### 4. Preserve structured fallback evidence

Do not discard `ProtocolError` in `resolve_download`. Emit a bounded, non-secret reason/stage for expiry, connect, carrier, stall, malformed resume, digest, spool, and install failures, with endpoint/extent counts, completed bytes, reconnect attempts, and elapsed time. Never log the opaque ticket or bearer material.

Please also expose the new concurrency/body/stall knobs through the normal Heddle user-config/env path; the current PR adds only programmatic `ClientConfig` fields/builders.

### Acceptance boundary

The result must remain one logical pull, one client authorization signature, automatically managed provider connections, and ordinary Weft fallback on the existing pull. The Worker remains primary for eligible immutable extents; Weft remains fallback and serves ineligible objects.

forge-zephyr:
🤖 Starting. Worktree `workspace/issue-1166`, branch `task/1166-provider-pull-bounded-concurrent-fan-out`. File scope: (inferred from epic; declare in PR).

forge-zephyr:
♻️ Released by orchestrator. Reason: manual release. Worktrees preserved for inspection: `workspace/issue-1166`.

Issue #95, opened 2026-05-17
refs_storage.rs lock file: drop file handle before Windows cleanup (same shape as heddle#86 r2)


thethorne48:
## Context

Surfaced during heddle#86 r2 Rule-7 sweep ([PR #94 commit \`38e3552\`](https://github.com/HeddleCo/heddle/pull/94/commits/38e3552)). Same Windows file-handle-vs-cleanup bug as the original heddle#86, in:

\`crates/refs/src/refs/refs_storage.rs:153-163\`

The lock-file write path:
1. Opens \`file\` with \`create_new\`
2. On \`write_all\` failure (line 158) calls \`fs::remove_file(&path)\` while \`file\` is **still in scope**
3. Same on \`sync_all\` failure (line 162)

On Windows this fails with \`ERROR_SHARING_VIOLATION\` and the lock file leaks. POSIX cleans up correctly because \`unlink\` on an open file succeeds.

## Why it matters

Daily heddle use on Windows where ref-storage write fails (disk full, network drop) leaks a lock file. The stale-lock detection at line 147 partially compensates by reaping locks older than 5 min, but until that runs, the next concurrent operation waits the full 5 min. For an active dev workflow, that's a 5-minute UX cliff after any IO blip.

## Fix

Apply the same pattern from heddle#86 r2:
1. Extract the write path into a helper that takes the file by value (ownership transfer)
2. Drop the file before \`remove_file\` in the error branch
3. Add a `DropTrackingFailingWriter`-style test to assert the ownership-transfer guarantee

Reference: heddle#86 r2's `write_body_or_cleanup<W: Write>` signature in `shared_target.rs` (commit \`38e3552\`).

## Acceptance criteria

- [ ] **Red-commit**: `DropTrackingFailingWriter` test (or equivalent) asserting the file handle is dropped before \`remove_file\` in both the \`write_all\` and \`sync_all\` error paths.
- [ ] Refactor the lock-file write into a helper with ownership-transfer semantics matching heddle#86 r2.
- [ ] Sweep again — any other refs/oplog/objects sites with the same shape that the heddle#86 + heddle#80 sweep didn't catch.

## Related

- [heddle#86](https://github.com/HeddleCo/heddle/issues/86) — original shared_target.rs fix + Windows-bug discovery (r2)
- [heddle#80](https://github.com/HeddleCo/heddle/issues/80) r3 — original init.rs fix (already safe per heddle#86 r2 sweep)

## Definition of Done

[...]

PR #1262, opened 2026-08-09
Audit legacy code paths and improve performance coverage

thethorne48:
## Summary

- Remove or simplify legacy code paths across CLI, repository, object storage, oplog, merge, semantic analysis, and wire layers.
- Improve fsck, sync, review, history, workflow, and Git overlay behavior.
- Add regression and performance coverage for core functionality, error paths, interoperability, multi-agent workflows, and oplog recovery.
- Update stability, legacy-removal, testing, command, and segmented-oplog documentation.
- Add Cargo audit configuration and refresh dependency metadata.

## Testing

- Expanded CLI integration, comprehensive, core-functionality, and crate-level regression tests.
- Added performance and interoperability coverage, including Git overlay and sync scenarios.
- Formatting, linting, and the relevant Rust test suites were run.

---
<a href="https://app.blacksmith.sh/HeddleCo/codesmith/heddle/pr/1262"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1788830967&installation_model_id=21489&pr_number=1262&repository=HeddleCo%2Fheddle&return_to=https%3A%2F%2Fgithub.com%2FHeddleCo%2Fheddle%2Fpull%2F1262&signature=02eeeab6285895d0e46ad4ffa8c9b1beea51fe0d8bfb15bdf077b063f7d0d2e9"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


forge-zephyr (changes requested):
## Review: request-changes

Multi-agent adversarial review across API-capabilities, wire, semantic/merge/object-storage/fsck, oplog, and test-quality/docs. The deletion sweep is **cleaner than its size** — nearly every deleted symbol is genuinely dead or had callers migrated, on-disk V4 oplog format is untouched, and fsck/perf tests got *stronger*. The risk is concentrated in **unflagged behavior/perf regressions and two data-safety issues smuggled into one squashed "cleanup" commit**, plus a cross-repo break invisible to this repo's CI.

### BLOCKER
- **`api-capabilities/heddle.json` lies about 3 shipped RPCs** (`GetReviewPayload` :2188, `ListSignatures`, `SignState`). The review-service de-protoization renamed `LocalStateReview` methods (`get_review_payload`→`payload`, etc.); `verify_sync.py` requires a structural binding in `crates/cli/src/cli/commands/review.rs` and finds none. It's **3 failures, not 1** (verifier exits on first). **Fix: rename the three methods back to the RPC-canonical names** (`get_review_payload`/`list_signatures`/`sign_state`) + update call sites — zero manifest churn. (Changing status/evidence won't pass: status is hash-pinned cross-repo, and the verifier rejects `.payload(` as not-RPC-bound.)

### HIGH
1. **Cross-repo silent break — deleted wire types are live in weft.** `message_refs.rs` removes `pub HeadInfo`/`RefsList`/`RefFilter`, but weft imports all three (`weft-hosted/src/service/refs_hosted_impl.rs:21-23`, `canonical_repository.rs:24,1853` — touched by weft#1142 *yesterday*). `heddle-wire` is a published crate weft pins at `0.11`; this PR doesn't bump the version, so a `0.11.x` publish is a semver-major break that lands in weft via `cargo update`, **outside heddle CI**. Fix: keep the three types until weft migrates, or bump to `0.12` with a coordinated weft migration.
2. **Oplog recovery permanently destroys intact committed segments.** `prune_invalid_suffix` (`segmented_oplog.rs:643`) + `sweep_unlisted_containers` (:549) `fs::remove_file` every segment after a salvaged-short container — only the *damaged* container is quarantined; the intact suffix (committed entries) is unlinked with no copy. Worse, this runs on the **read path** (`validate_current` via `open_index_for_read`), not just explicit `recover`. Fix: move pruned containers to the quarantine namespace instead of unlinking.
3. **Every repo open now reads+validates the entire oplog and discards it.** `open_raw` went from header-only O(1) (`validate_current_format`) to `ensure_valid_current_format`→full `read_to_end` of base + every segment (`repository.rs:680`), then throws it away; first oplog read re-opens everything. O(total oplog bytes) per command, twice — a per-command regression in a perf PR. Fix: cheap `is_healthy` probe at open, defer strict validation, or reuse the validated view as the cache.
4. **`EXTRACTOR_VERSION` not bumped despite a taxonomy change.** `symbol_resolver.rs:367` adds a `pair` arm to `visit_definitions` (the `extract_semantic_file` walker); unchanged JS/TS `{foo: () => {}}` now yields extra symbols + a different `scaffold_hash`, but `EXTRACTOR_VERSION` stays `3` (`semantic_index.rs:43`) → mixed/corrupt index, spurious semantic-hash diffs. Fix: bump to 4.
5. **LRU eviction became O(n) on a 65,536-entry cache.** `fs_store.rs:~205` `evict_to_fit` now does `entries.iter().min_by_key(...)` full-scan instead of `pop_front`; once full, every insert scans 65k entries → minutes of eviction scanning materializing large repos. Fix: intrusive LRU list or sampled eviction.
6. **FsStore read ordering inverted native-first → overlay-first** (`fs_impl.rs`, all get_/has_ methods), unexplained, in a "remove obsolete code" commit; native-only reads now pay a sqlite+git probe first, and with `external_source` set the zero-copy pack path is bypassed. Keep native-first + the (good) read-through cache.
7. **Deleted the only e2e tests for live embargo/out-of-band push reconciliation.** `git_projection_engine_integration.rs` (~10 tests) covered `push_network_remote` + forced-embargo-rewind — code still live in `git_core.rs` but now with **zero e2e coverage and zero prod callers**. A regression that re-advertises an embargoed commit on a real remote (privacy-leak class) would stay green. Fix: port the tests, or delete the now-dead push surface.

### Notable MEDIUM
- Downgrade/mixed-version oplog hazard with **no repo-format gate** — an old binary silently reads the stale `oplog.bin` base (silent truncation). Bump repo format or stub `oplog.bin` when the first manifest publishes.
- `heddle oplog recover` is unreachable exactly when needed (`open_repo` strict-validates first); a pure count/head manifest disagreement has no repair verb (`reconcile_manifest_metadata` exists but isn't invoked).
- Reserved verdict-envelope prefix guard dropped in `execute_sign_state` (`state_review.rs:362`) — a locally-signed prefixed justification syncs as a blob and can reach weft's verdict readers. Reinstate the `if` guard (defense-in-depth).
- Replay dedup encoding changed prost→serde_json under the same verb (`local_review/mod.rs`) — persisted-format change; ack or bump the verb to `.v2`.
- `commit_graph.rs` refactor collides semantically with the just-merged #1259 (bounded merge-base) — **rebase and re-run perf-contract before merge**.
- Diff + merge rename output can change (both are latent-bug fixes, `diff/mod.rs:1991`, `rename_matcher.rs:236`) — pin the fixed scenarios with a test; note in changelog.
- Overlay `NotFound` downgraded to silent miss + `SleyRepository` cached forever (long-running daemon won't see a later `git fetch`, silently).
- `promote_to_loose_uncompressed` guard inverted for both-present blobs (adopted repos lose loose-promotion/hardlink materialization).
- Transport coverage narrowed (`git://`/plain-HTTP/**basic-auth**/reserved-remote-name boundary tests gone with the legacy pull).
- Stack-overflow guard test runs at 5,000 commits vs the ~80k observed overflow (green-on-nothing) — run the negative case.
- `heddle watch` notify path is dead under V5 (watches `oplog.bin`, which V5 deletes) — degrades to polling. Watch the `oplog/` dir.

### Clean / positive (verified)
Wire deletions are hand-rolled dups now served by generated proto types; `symbol_extraction` fully reimplemented; **fsck got stronger** (new `missing_tree` + iterative cycle detection replacing silent skips); perf budgets unchanged with timed closures now `.expect()`-ing; V4 oplog on-disk format + durability ordering (tmp→fsync→rename→fsync-dir) untouched; docs/audit-config honest (no new advisories ignored). An assertion-free success-theater fsck test was deleted (honesty win).

### Top actions before merge
1. Rename the 3 review methods back (unblocks CI).
2. Resolve the `heddle-wire` cross-repo break (keep types or 0.12 + weft migration).
3. Oplog: quarantine-don't-delete pruned segments; restore cheap repo-open validation.
4. Bump `EXTRACTOR_VERSION`; fix O(n) eviction; justify-or-revert external-first read ordering.
5. Port-or-delete the embargo push surface.
6. Rebase onto main (for #1259).
7. **Split the squashed commit** — right now a bad piece can't be reverted without taking the good ones.

_Review by multi-agent analysis; findings verified against `8113b3a..codex/audit-legacy-code-and-performance`. Nothing was built — CI remains the gate for the pending Quality/Postgres/Release-contract jobs._

forge-zephyr:
## Fable review — full detail: **wire + state_review** (supplements the consolidated review)

### CI failure root cause (CONFIRMED)
The PR replaced proto-typed `LocalStateReviewService` with native-typed `LocalStateReview` (`crates/daemon/src/local_review/state_review.rs`), renaming `get_review_payload`→`payload`, `list_signatures`→`signatures`, `sign_state`→`sign`, and rewrote `crates/cli/src/cli/commands/review.rs` to call the new names. No functionality lost (same `ReviewShowOutput` JSON). But `api-capabilities/heddle.json:2188` declares the `cli` layer of `StateReviewService/GetReviewPayload` `shipped` with evidence needle `get_review_payload` in review.rs; `verify_sync.py::_validate_rust_binding` needs a structural `async fn get_review_payload(` / `.get_review_payload(` there and finds none. **3 rows fail** (`GetReviewPayload`, `ListSignatures`, `SignState`) — CI shows only the first (exits on first). **Correct fix: rename the three `LocalStateReview` methods back to RPC-canonical names + update review.rs call sites — zero manifest churn.** (Status change breaks the snapshot hash pinned to `HeddleCo/api@726958d`; evidence-needle change is rejected as not-RPC-bound.)

### Ranked findings
1. **BLOCKER (CONFIRMED)** — the 3 broken api-capabilities bindings above.
2. **HIGH (CONFIRMED)** — `message_refs.rs` deletes `pub HeadInfo`/`RefsList`/`RefFilter`, live in weft (`refs_hosted_impl.rs:21-23`, `canonical_repository.rs:24,1853`, touched by weft#1142 on 2026-08-08). Weft pins `heddle-wire = \"0.11\"`; PR doesn't bump version → publishing `0.11.x` is a semver-major break landing in weft via `cargo update`, outside heddle CI. Fix: keep the three types until weft migrates, or bump to `0.12` with a coordinated weft migration. (All other deleted wire types are clean.)
3. **MEDIUM (mechanism CONFIRMED)** — `execute_sign_state` (`state_review.rs:362`) dropped the base guard rejecting justifications starting with `VERDICT_ENVELOPE_TAG`. Weft still decodes that prefix as a verdict envelope (`canonical_search_state_review.rs:500,816,885`); weft's own guard only covers its hosted SignState RPC, but a locally-signed prefixed justification syncs as a `ReviewSignaturesBlob` attachment and can reach weft's verdict readers. Fix: reinstate the reserved-prefix `if` (one const).
4. **MEDIUM (CONFIRMED; likely acceptable pre-users)** — `with_idempotency` (`local_review/mod.rs`) changed persisted dedup encoding prost→serde_json under the same verb `state_review.sign_state`, and the body-hash input from proto to JSON bytes. Old records fail to decode; same request retried hashes differently → loud `Conflict`. Ack or bump the verb to `.v2`.
5. **MEDIUM/LOW (looks intentional)** — `workflow.rs:1121` + `operator_loop.rs:94` switched pre-land/smart sync from `GitProjection::pull` (mirror, AllRefs, up-to-date short-circuit) to `pull_current_git_overlay_authoritative` (current branch + notes only). Other remote branches no longer imported during sync/land; message `commits_seen`→`commits_imported`. Forced (pull removed, git_core.rs −518); new path has strengthened coverage. Confirm the fetch-scope narrowing is deliberate.

### Clean / nits
Wire has no envelope enum; deleted types are standalone name-keyed serde payloads — no discriminant reuse/renumbering; retained enums/fields untouched; `SymbolKindTag::tag_byte` (persisted) untouched. `SymbolKind::FunctionSignatureOnly` was dead. `signing_payload` unchanged → persisted signatures still verify. Dropped `record_verdict`/`record_check_ack`/`get_review_progress` had zero callers. `hosted/sync.rs` always-spool-to-disk is deliberate memory hardening.

forge-zephyr:
## Fable review — full detail: **oplog subsystem**

**Bottom line: no format-compat blocker** — V4 container encoding (`LMOPLOG\0`, container v4, record schema 4) is byte-for-byte untouched; `SegmentedOpLogIndex::open` adopts a manifest-less `oplog.bin` as generation-0 base (`segmented_oplog.rs:53-77`), so old data still reads. On-disk change is additive (`oplog.manifest` `HDOPV5\0\0` v2 + `oplog.segments/`, published from first post-upgrade append). Durability ordering correct (segments `write_file_atomic` temp→fsync→rename→fsync-dir, validated before manifest swap; manifest swap atomic); lock discipline held; manifest paths traversal-guarded; ≤64 binary-tier bound holds at append+parse. Findings:

1. **HIGH (borderline blocker) — recovery permanently destroys intact committed segments without quarantine.** `prune_invalid_suffix` (`segmented_oplog.rs:643`) drops every segment after a salvaged-short container, then `sweep_unlisted_containers` (:549, called at :184 and :482) `fs::remove_file`s them. Only the *damaged* container is `.corrupt`-quarantined (`packed_oplog.rs:1168-1179`); intact suffix segments (committed entries) are unlinked with no copy. Runs on the **read path** (`validate_current` via `open_index_for_read`, `oplog_core.rs:223-235`), not just `recover`. Test `recovery_prunes_suffix_after_a_middle_segment_loses_its_entry` asserts the deletion as desired. Fix: move pruned containers to the quarantine namespace.
2. **HIGH — every repo open fully reads+validates the entire oplog and discards it.** Base `open_raw`→`validate_current_format` (header-only, O(1)). Branch `open_raw`→`ensure_valid_current_format`→`SegmentedOpLogIndex::open`→`read_to_end` of base + every segment + full index validation (`repository.rs:680`), then thrown away; first oplog read re-opens via `load_cached`. O(total oplog bytes) per command, twice if it touches the oplog. Fix: cheap `is_healthy` probe at open, defer strict validation, or reuse the validated view as the cache.
3. **HIGH — downgrade/mixed-version hazard, no repo-format gate.** After first post-upgrade append, `oplog.bin` remains as a stale immutable base under a manifest the old binary doesn't know; old binary silently reads only the stale base (silent truncation), and if it appends, base count/head diverge → new binary fails forever with \"manifest disagrees with immutable segments\" (unrepaired, see #4). Repo format stays v3. Fix: bump repo format when the first manifest publishes, or replace `oplog.bin` with a refusal stub.
4. **MEDIUM-HIGH — `heddle oplog recover` unreachable when needed; metadata-only disagreement has no repair verb.** `cmd_oplog_recover` calls `open_repo()` first (`oplog.rs:96`) which strict-validates (finding #2) and fails for corrupt manifest / missing segment / metadata disagreement — so the operator entrypoint never runs for non-auto-repairable damage. And `validate_current` (:160-189) only runs `reconcile_manifest_metadata` when a container salvage occurred; a pure count/head disagreement returns `Ok` without fixing, then strict validation fails again → permanent brick despite `reconcile_manifest_metadata` (:627) existing for exactly that.
5. **MEDIUM — recovery report undercounts loss.** `recover` (`segmented_oplog.rs:459-474`) surfaces only the first repaired container's report; entries destroyed by suffix pruning (#1) never added to `entries_lost`.
6. **MEDIUM — `heddle watch` notify path dead under V5.** `tail_loop` watches the `oplog.bin` inode (`watch.rs:272-291`); V5 appends touch `oplog.manifest`/`oplog.segments/`, and compaction deletes `oplog.bin` (`segmented_oplog.rs:583-585`). Works only via poll-interval fallback. Fix: watch the `oplog/` dir.
7. **MEDIUM — one weakened assertion.** `truncated_latest_oplog_is_quarantined_and_salvaged...`: \"appendable afterward\" became `assert!(head_id() <= 3)` — vacuous (prior line asserts `== expected_count`). Mitigated by the CLI end-to-end salvage test.
8. **NIT — doc/ADR drift.** `oplog_backend.rs:~150` claims CoW-clone reuse (all reflink code deleted); ADR-0050 says `oplog.bin` retained after compaction (it's deleted by sweep); `commands_oplog.rs:12-13` help says recovery \"rebuilds oplog.bin\".

**Deleted-function check: clean** — `append_entries*`, `write_header`, `write_index_sections`, reflink machinery gone; all remaining callers go through `SegmentedOpLogIndex`; `#[cfg(test)]`-demoted fns have no non-test callers.

forge-zephyr:
## Fable review — full detail: **semantic / merge / object-storage / fsck / object-model / repo-history**

1. **HIGH (CONFIRMED) — `EXTRACTOR_VERSION` not bumped despite an extraction-taxonomy change.** `symbol_resolver.rs:367` adds a `pair` arm to `visit_definitions`, which is also the walker behind `extract_semantic_file` (`semantic_index.rs:125`); JS/TS `{ foo: () => {} }` now extracts extra symbols + a different `scaffold_hash` for unchanged source, but `EXTRACTOR_VERSION` stays `3` (`semantic_index.rs:43`) whose doc says to bump on exactly this. → mixed index (old nodes missing pair symbols) + spurious semantic-hash change reports. Fix: bump to 4 + doc line. No test catches it.
2. **HIGH (CONFIRMED, perf) — LRU eviction O(n) per evicted entry on a 65,536-entry cache.** `fs_store.rs:~205` `evict_to_fit` now `entries.iter().min_by_key(...)` full-scan instead of `pop_front`; `VERIFIED_LOOSE_BLOB_CACHE_CAPACITY=65_536` → once full, every insert scans 65k → minutes of eviction scanning materializing millions of blobs. Read-lock hit path is a genuine win; write path regressed from O(1). New test does 1000 reads + 1 insert → never exposes it. Fix: intrusive LRU list or clock/sampled eviction.
3. **HIGH (CONFIRMED behavior change) — external-source authority inversion in FsStore.** `fs_impl.rs` — `get_blob(_bytes)`, `blob_size`, `has_blob`, `get_tree(_serialized)`, `has_tree`, `get_state`, `has_state`, `list_states` now consult the Git overlay **before** native (was native-first). Content still correct, but native-only reads pay a sqlite `sha_map` query + git probe first; with `external_source` set, `get_blob_bytes` (:712) / `get_tree_serialized` (:345) unconditionally bypass the zero-copy pack path for all objects. No rationale in the squashed commit; the (good) read-through cache didn't require external-first. Fix: keep native-first + read-through cache on fallback.
4. **MEDIUM (CONFIRMED) — `promote_to_loose_uncompressed` guard inverted for both-present blobs.** `fs_impl.rs:~914` returns `Ok(false)` whenever the overlay has the blob, even when a native copy exists → adopted/projected repos lose loose-promotion + hardlink-first materialization (falls back to byte copies). Also `has_blob_locally`'s probe lost its in-memory-cache fast path (non-overlay stores now always stat/pack-probe).
5. **MEDIUM (CONFIRMED + PLAUSIBLE) — overlay `NotFound` downgraded to silent miss; cached `SleyRepository` staleness.** `git_overlay_object_source.rs:127-131,153-157` — a hash mapped in `sha_map` but whose git object is gone now returns `Ok(None)` (was a hard error) → readers treat corruption as absence (fsck still flags, normal reads degrade silently). `SleyRepository` cached forever per store (`git()` :219-235); if sley doesn't refresh its pack list on miss, a long-running daemon won't see objects from a later `git fetch` — silent with the new None. (Cached sqlite conn is fine.)
6. **MEDIUM (CONFIRMED, bug fix w/ output change) — core/diff rename output can change.** `diff/mod.rs:1991-2102` — old greedy assignment had a short-circuit bug (`used_old.insert(a) && used_new.insert(b)` marked `a` used even when `b` taken, dropping `a`'s remaining candidates). New `RenameCandidateIndex::assign` finds them + deterministic (score, lex-path) tie-break. User-visible diff change shipped as refactor; no changelog note, no pinning test.
7. **MEDIUM (CONFIRMED, improvement w/ output change) — merge rename tie-breaking was nondeterministic; now fixed.** `rename_matcher.rs:236` — `FlatTree` is a HashMap so old candidate order (hence tied greedy assignment) varied run-to-run; branch sorts by path + tie-breaks (score desc, source_index, target_index). Results stable but may differ from a prior run. Nit: `RenameCandidateIndex::push` panics on non-finite score (unreachable today, but a panic seam in a now-`pub` API).
8. **LOW — semantic symbol resolution broadened.** `find_definitions` now filters `visit_definitions` (`symbol_resolver.rs:1017-1034`) → resolves consts/statics/modules/interfaces/type-aliases/Zig/C/Java the old walker ignored; changes blame/review symbol-target behavior; only extraction tests cover the new kinds, not the resolver-with-parent-filter path.

**Verified clean (highest-risk areas):** deleted `symbol_extraction.rs` fully reimplemented (zero remaining refs; `DefinitionKind` now re-exports `SymbolKindTag`); **on-disk format unchanged** (layout/naming/hashing/atomic-ordering intact — old stores read fine); **fsck stronger not weaker** (new `missing_tree` + missing-provenance-subtree detections replacing silent skips; iterative cycle detection with 50k-deep test; positive assertions on corrupted fixtures); `repository_history`/`commit_graph` traversal semantics unchanged (rewrite batches cache persistence into one `finish()` + hoists full-closure load to session start); `detect_file_renames` semantics equivalent to base, now O(files) not O(pairs). 

**Bottom line:** no blocker here, but #1 (EXTRACTOR_VERSION), #2 (O(n) eviction), #3 (external-first ordering) should be fixed before merge — #3 needs explicit author rationale (significant un-flagged behavior/perf change inside a single \"refactor + remove obsolete code\" commit).

forge-zephyr:
## Fable review — full detail: **test quality + docs truth**

1. **HIGH (CONFIRMED) — deleted the only e2e tests for live wire-remote push reconciliation (embargo / out-of-band safety).** `crates/cli/tests/git_projection_engine_integration.rs`: ~10 tests deleted whose code is still live in `git-projection/src/git_core.rs` — `push` (:932), `push_with_scope(_force)` (:943,:963), `push_network_remote` (:4147), forced-embargo-rewind / served-frontier (:3090-3293). Deleted: `retraction_delete_propagates_to_url_remote`, `foreign_ref_on_url_remote_survives`, `out_of_band_destination_descendant_not_force_overwritten`, `out_of_band_destination_tag_not_overwritten`, `out_of_band_advance_after_embargo_not_deleted`, `heddle_published_tip_embargo_rewind_still_forced`, `scoped_push_propagates_cross_thread_embargo_to_destination`, `push_exports_local_branches_and_tags_to_path_remote`, `push_current_thread_scope_exports_only_attached_branch_to_path_remote`, `scoped_export_preserves_unminted_out_of_scope_public_tag`, + the `GitDaemon`/`GitHttpBackend` harness (whose doc said it was \"the only way to cover URL/network destination reconciliation, heddle#316 r11\"). Compounding: those push entry points now have **no in-repo production callers** either. Failure: `push_network_remote` regresses (drops force flag on rewrite) → embargoed commit stays advertised on a real remote → privacy-leak class, zero test to catch. Fix: port the url-remote tests to the surviving harness, or delete the dead push surface (after checking weft consumers of the crate API).
2. **MEDIUM (CONFIRMED) — legacy pull deleted cleanly, transport coverage narrowed.** `GitProjection::pull`/`fetch` deleted with their tests; callers migrated to `pull_current_git_overlay_authoritative`, tested over local path + HTTPS+private-CA — but `git://`, plain-HTTP, and **basic-auth HTTP** cases have no equivalents, and `fetch/pull_rejects_reserved_git_remote_name_at_boundary` are gone though `reject_reserved_git_remote_name` is still live (`git_core.rs:188`). Fix: confirm the supported transport matrix; add basic-auth + reserved-name boundary tests at the surviving call sites.
3. **MEDIUM (PLAUSIBLE) — deep-history stack-overflow guard may be green on the code it claims to guard.** New `import_handles_deep_linear_history_without_stack_overflow` builds 5,000 commits, but its own comment says recursion overflowed at ~80k on git/git — 16× below the observed point. Classic \"green is not working.\" Fix: run the negative case (revert the iterative walker `walk_tree_iterative`, confirm it SIGABRTs at DEPTH=5000; if not, raise DEPTH or shrink the test-thread stack).
4. **CLEAN/POSITIVE (CONFIRMED) — fsck deletions removed no-op tests, replaced with stricter ones.** `test_fsck_reports_corrupted_object` ended `let _result = heddle(...)` (zero assertions); `test_fsck_detects_broken_parent` accepted every outcome. Both deleted; added real `missing_tree` finding (`fsck/objects.rs:56`), `test_tree_blob_checks_report_missing_references` (`fsck/tests.rs:258`), and `expect_err` + message match on pack corruption.
5. **CLEAN (CONFIRMED) — perf tests: no thresholds loosened, no new skips/no-op assertions.** All budgets unchanged (20s/1000-file snapshot, 5s/10s status, 2s/4s diff/log/blame). Every `let _ = heddle(...)` in a timed closure became `.expect(...)` (fixed a loose-perf antipattern). The one removed timing assertion (`credential.rs:696`) was replaced with a deterministic `oversized_helper_outputs` counter. The 1874-line deletion is the network-harness + legacy-pull removal (findings 1-2), not perf-test gutting.
6. **CLEAN (CONFIRMED) — docs truth + audit config.** `.cargo/audit.toml`/`deny.toml`: no new advisories ignored. STABILITY.md oplog claims match code (`HDOPV5\0\0` v2; `LMOPLOG\0` container v4). ADR-0050 \"≤64 append containers\" matches the u64 counter design, honest about cold-open cost. LEGACY_DELETION_NEXT_WAVE.md V5/V4 refusal claims match `packed_oplog.rs:46-57`. CHANGELOG \"obsolete protobuf-shaped wire messages gone\" verified (deleted types have no remaining Rust refs; surviving grep hits are RPC permission-name strings).
7. **NITS.** `oss_cli_polish.rs:8905` TTY test silently `return`s on non-Linux (still hard-required on Linux). `state_management.rs` harness injects the test principal — checked 3 refusal tests, no behavior masked. New regression spot-checks are genuinely stronger (error_paths duplicate-thread/binary-diff assert exact diagnostics; git_overlay_sync_adoption adds branch/tracking-ref/clean-worktree + idempotent-rerun proofs; oplog_salvage retargets to manifest-selected segments with quarantine assertions).

**Bottom line:** the perf/docs/audit half is honest and mostly *strengthens* tests. The one real problem is finding 1 — deleted integration tests guarded live embargo/out-of-band push-reconciliation code (leak-prevention surface) now with zero e2e coverage and zero prod callers: port the tests or delete the dead surface before merge.

forge-zephyr (dismissed):
Findings all addressed (blocker + 7 highs fixed/already-fixed in prior rounds), rebased onto main, and reconciled cleanly with #95's declaration (StateReview stays intentionally-unsupported, no shipped RPC reintroduced; verify_sync + 13 attestation invariants pass). The 4 documented deferrals are filed as follow-ups. Approving.

forge-zephyr:
Merge-integration fix pushed in `bdbb5f949773a6091dc87fd84ffc54db738233d8`.

**What the merge exposed**

The rebased Quality failure was an exit-101 test failure, not a Rust compile error. Under the Quality runner's four-way test execution, generic CLI integration subprocesses inherited the new/default `auto` fsmonitor behavior and concurrently spawned detached native helpers. Those helpers could outlive their temporary repositories or leave stale endpoints; local reproduction produced `helper_unavailable_no_full_scan`, `Connection refused`, and monitor warnings contaminating structured stderr. The reported empty/0% coverage was downstream of that failed test run; the current `Quality` job itself does not invoke `cargo llvm-cov`.

**Fix**

- Pin the Quality libtest concurrency to four threads, matching the 4-vCPU lane and the concurrency used to verify #1291.
- Make generic subprocess test harnesses disable ambient fsmonitor helper startup, so unrelated CLI tests cannot contend through background helpers.
- Keep the audit contract intact: the clone/fsmonitor tests explicitly opt their positive `status` checks into `HEDDLE_FSMONITOR=native`; their clone-must-not-start-helper assertions are unchanged.

No audit implementation or assertion was removed.

**Passing local gates** (clean clone under `/home/scratch`, isolated Cargo targets)

Quality exact commands:
- `cargo build --locked --workspace --tests --bin heddle --bin heddle-fuse-worker` — PASS
- `cargo clippy --locked --workspace --lib --bins --tests --examples -- -D warnings` — PASS
- `cargo test --locked --workspace -- --test-threads=4` — PASS (107 suites; 5,113 passed, 0 failed, 67 expected ignored)
- `python3 -m unittest discover -s scripts/tests -p "test_*.py" -v` — PASS (12/12)
- `bash scripts/check-default-cli-contracts.sh` — PASS (no schema/sample drift)

Feature contracts:
- `heddle-repo`: `git-overlay,zstd` — PASS
- `heddle-repo`: `native,zstd` — PASS
- `heddle-cli`: `git-overlay,client,semantic,zstd` — PASS
- `heddle-cli`: `native,semantic,zstd` — PASS

PR remains ready for review (not draft).

forge-zephyr:
Follow-up integration blocker after the fix push:

While the local Quality and Feature-contract gates were running, `main` advanced from `1691c3c7` to `639606d3` (five commits), including `5fdced40` / #1296 (`ci: split fast lane from full suite`). GitHub now reports this PR as conflicting, so it has not scheduled the Rust workflow for `bdbb5f94`.

A read-only merge check shows this is a semantic workflow conflict, not a routine line overlap:
- modify/delete conflicts: `.github/workflows/api-capability-sync.yml`, `audit.yml`, `release-pipeline-check.yml`, and `unwrap-inventory.yml` (this audit PR deletes them; #1296 modifies them)
- content conflicts: `.github/workflows/rust-tests.yml`, `perf-core-loop.yml`, and `publish-crates.yml`

Per the instruction to stop rather than guess on a design decision, I did not merge the newly advanced `main`. The isolated fsmonitor test fix remains committed and locally verified at `bdbb5f94`; reconciling whether #1296's workflow split supersedes or should coexist with this PR's workflow deletions needs an explicit decision before another rebase/merge and gate run.

forge-zephyr:
Closing — superseded. This monolithic legacy-audit PR was deliberately **broken into slices that have since merged individually** (tracked and completed as the "legacy-audit slices" work). The conflicts here are exactly because those slices already landed the same changes on main. Nothing unique remains to rebase; re-slice anything still wanted rather than reviving this grab-bag.

PR #1746, opened 2026-09-13
fix(netd): Heddle home relay via `preview` feature, never n0 Default

forge-zephyr:
## Summary

- `heddle netd serve` bound the persistent endpoint with `RelayMode::Default`, so iroh homed on n0 (`*.relay.n0.iroh.link`). Live log: `home is now relay https://usw1-1.relay.n0.iroh.link./`.
- Hosted CLI connections stay as-is: they use `RelayMode::custom(relays)` from the signed endpoint descriptor (source of truth for hosted).
- Netd `default_relay_mode()` is now `RelayMode::custom` with **exactly one** Heddle home relay, selected at compile time by the **`preview` cargo feature** (not `debug_assertions`):
  - Default / `--release` (feature off) → `https://relay.heddle.sh/` only
  - `--features preview` → `https://relay.preview.heddle.sh/` only
- `heddle-cli` forwards `preview = ["hosted-client/preview"]`. The unused URL is cfg'd out, so a production binary cannot embed preview and a preview binary cannot embed prod. Never `RelayMode::Default`.

## Preview AX build

```
cargo build --release -p heddle-cli --features client,preview
```

Stock / crates.io / tagged release (prod home relay):

```
cargo build --release -p heddle-cli --features client
```

(`client` is already in the CLI default feature set; listing it is explicit.)

## Closes

Refs HeddleCo/heddle#1533
Refs HeddleCo/heddle#1620

## Red commit
`cd6cd652e78e50b1376d40bf524c17c355037d59`

## Test evidence

Debug, feature off (prod only):

```
$ cargo test --locked -p heddle-hosted-client --features client --lib default_relay_mode -- --nocapture
    Finished `test` profile [unoptimized + debuginfo] target(s) in 10.37s
     Running unittests src/lib.rs (target/debug/deps/hosted_client-31b582925b429dc4)

running 1 test
test network::tests::default_relay_mode_is_heddle_custom_for_this_build_flavor ... ok

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 463 filtered out; finished in 0.00s
```

Debug, `--features client,preview` (preview only):

```
$ cargo test --locked -p heddle-hosted-client --features client,preview --lib default_relay_mode -- --nocapture
    Finished `test` profile [unoptimized + debuginfo] target(s) in 15.90s
     Running unittests src/lib.rs (target/debug/deps/hosted_client-bb9e7d743b4f10c0)

running 1 test
[...]

PR #1747, opened 2026-09-13
fix(hosted): drop the 1s close tax and reuse weft through netd

forge-zephyr:
## Summary

- Preview `whoami` / push / pull / clone sat at **~2.0s wall even when warm-repeated**. The last **~1000ms** was `HostedConnection::close` awaiting `Router::shutdown` → `Endpoint::close` → `noq wait_all_draining` after the RPC was already `LocallyClosed`.
- One-shot CLI close now **starts** graceful shutdown (so we do not drop the endpoint dirty) and **bounds the foreground wait at 20ms**. On timeout the `JoinHandle` is **`mem::forget`’d** so drain keeps running. Wrapping the handle in `tokio::time::timeout` dropped it on `Elapsed` and cancelled mid-close — the opposite of detach.
- When `heddle netd serve` is running, hosted verbs reuse netd’s persistent endpoint and a **cached weft QUIC connection** over a same-uid UDS bridge. The CLI still signs every call; netd never sees credentials. Relays for those sessions still come from the **signed descriptor**. Preview vs prod home-relay stays the `preview` cargo feature from #1746 — this does not stuff both maps into one `Custom` set.
- Provider (CAS) dials on a netd-proxied session no longer fail closed. `ProviderBackend` still needs a real Iroh `Connection`, which cannot cross the UDS splice, so the CLI binds **one ephemeral local endpoint per process** the first time a provider is used. Weft stays in netd. Close of that local endpoint is the same 20ms-bounded drain.

## Closes

Refs HeddleCo/heddle#1746

## Measured 2s split (preview, warm-repeated whoami)

| Phase | ~ms |
|------|----:|
| HTTPS well-known descriptor | 130 |
| Cold relay WSS/TLS | 420 |
| Connection established (relay path) | 661 |
| whoami RPC until LocallyClosed | +330 → ~993 |
| **Idle until relay Shutdown / command exit** | **+1000 → ~1993** |

Local capture/context/discuss stay ~8–23ms. Target for warm networked ops is tens of ms once netd is up.

## Before / after

**Without netd (still the default one-shot path):**
- Teardown gap should collapse from ~1000ms to ≤20ms foreground.
- Residual floor is still cold bind + descriptor + relay + QUIC + RPC (~0.7–1.1s on preview). That is honest: a new process cannot reuse a connection it does not have.

**With `heddle netd serve` already running (warm reuse):**
- Second `whoami` should skip HTTPS descriptor fetch, relay WSS, and a new QUIC session.
- Residual floor: CLI process start + UDS Ensure/OpenBi + WhoAmI on the cached connection + bounded close (no-op on the proxied weft handle). Target: tens of ms once warm. Descriptor refetch only happens if the cached session expired or the QUIC connection died.
- `pull` / `clone` that need providers still pay a **per-process** local bind + provider WSS on the first CAS dial. Subsequent extents in that process reuse the cached provider connection. Cross-process provider reuse is not in this change (weft is).

## How teardown changed

`HostedConnection::close` still sends `CONNECTION_CLOSE`, then `tokio::spawn`s the shutdown. Foreground `select!` races the join against `FOREGROUND_ENDPOINT_DRAIN` (20ms). If the bound wins, `std::mem::forget(task)` detaches the handle so close is not cancelled. A proxied/netd handle does not close the daemon weft session. If that process bound a local provider endpoint, close starts the same bounded drain on it.

## How reuse works (netd, not a second daemon)

`heddle netd serve` already held the persistent device endpoint. It now also binds `<heddle_home>/state/heddle-netd-hosted.sock`. `HostedSession::connect` tries that bridge first (`Ensure` then per-RPC `OpenBi` spliced as raw stream bytes). Cache key is the server authority; a live, unexpired weft connection is reused. If the socket is missing, the CLI falls back to the local bind path with the new bounded close.

No auto-start: warm reuse is available when netd is running. Claim-bridge and preview/prod feature split are unchanged.

## Content durability across bounded close

`push_content_survives_bounded_close_and_clones_byte_for_byte` pins the weft-fetches-synchronously assumption:

1. Stage the native pack for the pushed closure (unpublished).
2. Complete the framed Push; the fixture publishes that pack only after the client stream has drained (`PushComplete.success`).
3. Hold shutdown 80ms so `close()` hits the 20ms bound and `mem::forget`s the drain.
4. Drop that client/endpoint. Connect a **fresh** client to the same server address.
5. Full clone; assert the payload blob is byte-for-byte `durable-across-bounded-close-e0c3\n`.

If Push never completes, the pack stays unpublished and the clone has nothing.

## Patch coverage (this revision)

codecov/patch was 75.89% with the gaps on the new netd/proxied paths. New tests drive those production paths instead of `#[allow]` or deleting code:

**`hosted_bridge.rs`**
- Cold Ensure / OpenBi fail closed (no cached weft → descriptor fetch, no invented session)
- Expired and closed weft cache eviction
- Provider OpenBi without a provider transport fails closed
- Missing socket, oversized frame, client disconnect before handshake
- Protocol mismatch: Opened-for-Ensure and Ready-for-OpenBi
- `connect_weft` on signed direct addresses, and fail-closed when direct + relays are gone

**`connection.rs` / `session.rs` / `mod.rs`**
- `connect_via_netd` missing-socket fail-closed
- Warm `connect_via_netd` + proxied `open_bi` (`write_all` / `write_chunk` / `finish` / `read_to_end` / `read_chunk` / `reset` / `stop`) and proxied close without a local provider
- `HostedClient::connect_via_netd` unary on proxied streams
- `HostedSession::connect` reuse when the bridge is up; fallback to local discovery when it is not (`connect` and `connect_outbound`)
- `connect_preferred` local fallback for `connect_with_config` and `connect_outbound_with_config` when netd is down

**`network.rs` / `net_endpoint.rs`**
- `bind_persistent_hosted` binds the persisted device endpoint

## Live verification (not CI proof)

On a preview-featured binary, against `https://api.preview.heddle.sh`:

```sh
# Cold / no netd — teardown gap should collapse
RUST_LOG=debug time heddle whoami --server api.preview.heddle.sh
# Repeat. Expect command_body_ms ~1s not ~2s; no 1s gap after LocallyClosed.

# Warm reuse
heddle netd serve   # leave running
RUST_LOG=debug time heddle whoami --server api.preview.heddle.sh   # first may still cold-fill netd
RUST_LOG=debug time heddle whoami --server api.preview.heddle.sh   # should drop far below ~2s
```

Look for `hosted connect using netd warm bridge reused=true` and the absence of a ~1s LocallyClosed → Shutdown gap. A `pull`/`clone` through netd should succeed (providers dial the local ephemeral endpoint) rather than fail with “opened as streams”.

## Red commit

N/A — fail-closed tests landed with the fix. They fail if close again blocks ≥~1s after LocallyClosed, if a second OpenBi on a warm netd session opens another QUIC connection, if a proxied session refuses provider dials instead of binding a local endpoint, if a drain that outlives the 20ms bound is aborted instead of completing after detach (`bounded_shutdown_still_completes_after_foreground_detach`), or if pushed blob bytes are missing after bounded close + a fresh clone (`push_content_survives_bounded_close_and_clones_byte_for_byte`).

## Test evidence

HEAD `8f32ed581cd1debcf489d30f2e43dc68c5efa908`

```
cargo clippy --locked -p heddle-hosted-client --lib --features client,test-utils -- -D warnings -D dead-code
Finished `dev` profile

cargo clippy --locked -p heddle-hosted-client --lib --bins --tests --examples --features client -- -D warnings -D dead-code
Finished `dev` profile

cargo test --locked -p heddle-hosted-client --lib --features client
test result: ok. 490 passed; 0 failed; 1 ignored
```

`full-ci` is on the PR. Leaving draft. Do not treat live preview AX or `pr-to-status` as CI proof.

## Manual verification

N/A — covered by tests. Live preview timing is documented above for a human/agent with preview credentials.

## Blocked by → resolved

None

<div><a href="https://cursor.com/agents/bc-e642ca4e-aab6-4696-8502-a288cc12e0c3?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-e642ca4e-aab6-4696-8502-a288cc12e0c3&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

---
<a href="https://app.blacksmith.sh/HeddleCo/codesmith/heddle/pr/1747"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1791862216&installation_model_id=21489&pr_number=1747&repository=HeddleCo%2Fheddle&return_to=https%3A%2F%2Fgithub.com%2FHeddleCo%2Fheddle%2Fpull%2F1747&signature=8869b6eb2cb6df8c655a0f534a27bf6919f671439e2742d948c1bd19592d90b4"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/crates/cli-args/src/cli/cli_args/commands_main.rs
+++ b/crates/cli-args/src/cli/cli_args/commands_main.rs
@@ -509,10 +509,12 @@ Examples:
     /// `heddle netd serve` runs a long-lived async daemon that binds
     /// the machine's single persistent Iroh endpoint on the persisted
     /// device node id and keeps it relay-reachable, so outstanding
-    /// claim links keep resolving across restarts. Unlike `daemon`, it
-    /// is not gated on Linux/FUSE and never idle-exits. `status`
-    /// reports liveness and the advertised node id; `stop` asks a
-    /// running daemon to close its endpoint and exit.
+    /// claim links keep resolving across restarts. Hosted verbs
+    /// (`whoami`, `push`, `pull`, `clone`) reuse that endpoint's warm
+    /// weft session when netd is running. Unlike `daemon`, it is not
+    /// gated on Linux/FUSE and never idle-exits. `status` reports
+    /// liveness and the advertised node id; `stop` asks a running
+    /// daemon to close its endpoint and exit.
     Netd {
         #[command(subcommand)]
         command: NetdCommands,
@@ -610,8 +612,8 @@ pub enum DaemonCommands {
 #[derive(Clone, Debug, clap::Subcommand)]
 pub enum NetdCommands {
     /// Run the foreground network daemon: bind the persistent device
-    /// endpoint, keep relays online, and serve same-uid control RPCs.
-    /// Never idle-exits.
+    /// endpoint, keep relays online, hold warm weft sessions for hosted
+    /// CLI verbs, and serve same-uid control RPCs. Never idle-exits.
     Serve,
 
     /// Report network-daemon liveness and the advertised device node
--- a/crates/cli-contract/src/cli/help.rs
+++ b/crates/cli-contract/src/cli/help.rs
@@ -650,6 +650,10 @@ const DAEMON_TOPIC: &str = "The mount daemon owns virtualized workspace sessions
 \n\
 `heddle daemon`        — FUSE mount-daemon control plane. Owns FUSE sessions for\n\
                          `--workspace virtualized --daemon` threads. Linux only.\n\
+                         Subcommands: serve | status | stop.\n\
+`heddle netd`          — Box network daemon. Binds the persistent Iroh endpoint\n\
+                         and holds warm weft connections so hosted whoami/push/\n\
+                         pull/clone reuse descriptor+relay+QUIC across commands.\n\
                          Subcommands: serve | status | stop.\n";
 
 const MODEL_TOPIC: &str = r#"Heddle mental model — the everyday loop in one screen.
--- a/crates/cli/src/cli/commands/netdaemon/mod.rs
+++ b/crates/cli/src/cli/commands/netdaemon/mod.rs
@@ -10,7 +10,8 @@
 //! Three verbs hang off this module, mirroring the mount daemon's
 //! control surface but box scoped:
 //!
-//! * `heddle netd serve` — foreground async daemon.
+//! * `heddle netd serve` — foreground async daemon (persistent
+//!   endpoint + warm weft sessions for hosted CLI verbs).
 //! * `heddle netd status` — liveness + advertised node id.
 //! * `heddle netd stop` — drain and exit.
 //!
--- a/crates/cli/src/cli/commands/netdaemon/server.rs
+++ b/crates/cli/src/cli/commands/netdaemon/server.rs
@@ -10,19 +10,21 @@
 //! lifecycles (this one never idle-exits) and different platforms
 //! (this one runs on any Unix).
 //!
-//! What it owns (heddle#1533, piece 1):
+//! What it owns (heddle#1533, piece 1, plus the hosted warm-session
+//! bridge):
 //!
 //! * one Iroh endpoint on the persisted device node id, bound with
 //!   relays online (browsers holding only a claim link dial through a
 //!   relay),
 //! * a box-scoped endpoint-discovery file advertising that node id,
 //! * a same-uid control socket for `netd status` / `netd stop`,
+//! * a same-uid hosted bridge that caches weft QUIC sessions so
+//!   one-shot CLI verbs reuse descriptor+relay+QUIC,
 //! * a single-writer guard so two processes never both bind the
 //!   device node id.
 //!
-//! What it does NOT own yet: the claim-ALPN router (piece 3 /
-//! heddle#1620) mounts on the endpoint at the seam marked below, and
-//! the weft subscription/doorbell (piece 2) is separate.
+//! The weft subscription/doorbell (piece 2) is separate. Preview vs
+//! prod home-relay remains the `preview` cargo feature.
 
 use std::{
     os::unix::net::UnixStream,
@@ -74,7 +76,7 @@ pub async fn run_network_daemon() -> Result<()> {
     // device node id, relays online. The node id therefore survives a
     // restart — the acceptance clause the browser claim link relies on
     // (heddle#1620).
-    let endpoint = hosted_client::network::bind_persistent_endpoint(
+    let (endpoint, hosted_sessions) = hosted_client::network::bind_persistent_hosted(
         hosted_client::network::default_relay_mode(),
     )
     .await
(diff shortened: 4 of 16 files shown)


forge-zephyr:
## Review — close/write-durability trace

Traced the close + push paths. **Write-durability is preserved and the bounded close is safe** — good work, and the `mem::forget`-detach vs `tokio::time::timeout`-cancel distinction is exactly right.

Why it's safe:

- **`close()` runs after the command's work, not during it.** Every call site is `client.close().await` as the final step of the handler (`clone.rs:1604/1955/2222`, `remote_ops.rs:1113`, `promote.rs:40`, …). Sequence is: RPCs + content transfer complete → *then* bounded close.
- **The 20ms bound only touches endpoint teardown, never data delivery.** `close()` sends `CONNECTION_CLOSE` then bounds only `router.shutdown()` (the close-frame-ACK drain). Unary RPCs prove receipt before returning — `unary_encoded` does `send.finish()?` then `recv.read_to_end(...)`, so the server has received+processed the full request before the call returns. A clipped teardown drain at worst drops the close-frame ACK (server sees a connection drop, not lost data) — cosmetic.
- The Proxied/netd path correctly does **not** drain netd's weft session on handle-drop (netd owns it); it only bounded-closes this process's lazy local provider endpoint.

**One coverage gap to close before merge** (not a correctness objection):

Bulk content bytes ride the provider (Iroh CAS) path, not these RPC streams. Content-push durability rests on weft fetching the referenced objects **synchronously within the push RPC** (so `read_to_end` on the push response implies content received). That's a cross-service assumption that no test in this PR pins. The added tests cover close-timing (`bounded_shutdown_*`, `close_does_not_block_*`) and warm-reuse, but there's no **push→clone round-trip asserting content is durable across the bounded close**. Please add one: push content, bounded-close + drop the endpoint, then clone from a fresh process/endpoint and assert byte-for-byte content — ideally with the drain forced to hit the 20ms bound so it exercises the detach path. That converts "safe by the weft-fetches-synchronously assumption" into "safe by test."

Also: only `fast-check` has run (this is a draft) — please add the **`full-ci`** label so `Check & test` / `Coverage` / the full suite exercise the QUIC/teardown changes before merge. (The `pr-to-status` red is board-sync noise, not code.)

forge-zephyr:
## full-ci is red — one mechanical fix

Held the merge: `full-ci` went red (fast-check / Check & test / Quality / Feature contracts / full-suite), and it's a real build break, not the `pr-to-status` board-sync noise. Root cause:

```
error: methods `pull_fixture` and `stage` are never used
  --> crates/hosted-client/src/hosted_runtime/hosted/test_server.rs:335
error: could not compile `heddle-hosted-client` (lib) due to 1 previous error
```

The new durability scaffolding — `PullFixture` (:321), `DurableSyncStore` (:329) + its `pull_fixture`/`stage`, and `start_durable_push_clone` — isn't `#[cfg(test)]`-gated, unlike every sibling fixture at :123–:306. CI's clippy `-D warnings` compiles the **lib target without tests**, so these are dead code there and the build fails. Your local `cargo test --lib --features client` compiled *with* the tests, which is why it was green locally — classic green-locally / red-in-CI.

Fix: put `#[cfg(test)]` on the new `DurableSyncStore`/`PullFixture` types and the `start_durable_push_clone`/helper fns (match the sibling fixtures at :123–:306). Then it'll pass the all-targets clippy lib build. Once full-ci is green I'll merge.

thethorne48:
@codex review

PR #1780, opened 2026-09-21
Remove hosted repository dependencies from local crates

forge-zephyr:
## Summary
- Export hosted thread, dedup, signing, ancestry, and semantic assembly APIs from leaf crates
- Make semantic assembly read-only until an explicit persistence method and expose typed cancellation, deadline, source-limit, output-limit, malformed-object, and storage failures
- Add bounded async ancestry with caller-owned cancellation/deadline and leaf-level behavior tests
- Adopt Sley 0.11 and select remote/worktree/hooks/history-editing/TLS features only in local consumers
- Add an isolated hosted-consumer graph gate excluding heddle-repo, SQLite, Ureq, sley-remote, worktree, hooks, and sequencer
- Serialize CLI unit tests to avoid process-global HOME/HEDDLE_HOME races

## Testing
- `cargo test --locked --workspace --lib --bins -- --test-threads=1` — passed locally (complete configured workspace library/binary suite)
- `cargo check --locked --workspace --all-targets` — passed
- `cargo clippy --locked --workspace --all-targets -- -D warnings -D dead-code` — passed
- `cargo test -p heddle-semantic --lib index_assembly::tests` — 4 passed
- `cargo test -p heddle-objects --features async-source --lib async_ancestry_tests` — 1 passed
- `cargo test -p heddle-repo --features async-source --lib async_ancestry_matches_sync_commit_graph_on_fixture` — 1 passed
- `bash scripts/check-hosted-leaf-consumer.sh` — passed; forbidden hosted dependencies absent
- `cargo fmt --all -- --check` and `git diff --check` — passed


forge-zephyr:
The extraction is useful, but three upstream gaps still prevent Weft from completing the dependency cleanup. Checked against `438454c` (the release follow-up is included).

- [ ] **Narrow the shared Sley dependency.** `Cargo.toml` still enables `remote` and `history-editing` in `[workspace.dependencies]`, and `heddle-objects` inherits that selection. Move local-operation features onto the actual local consumers. After adopting HeddleCo/sley#241, shared server-used crates must not enable worktree/hooks/sequencer through another parent. Local HTTP callers that use the built-in client also need to select the appropriate new client/TLS features explicitly.
- [ ] **Export the async ancestry helper from a leaf crate.** The new `objects::transfer::is_ancestor_from_source` accepts synchronous `ObjectSource`; Weft's hosted refs path uses `repo::is_ancestor_async` with `AsyncObjectSource`. Move that async API and the helpers it needs below `heddle-repo`, preserving its behavior and parity tests.
- [ ] **Complete the dedup extraction.** `DedupOutcome` moved, but `hash_request_body` and `DEFAULT_RETENTION_SECS` remain in `repo::operation_dedup`. Weft imports all three. Export the shared hashing and retention contract alongside the extracted vocabulary.

Please add an isolated consumer build/graph check for the server-facing configuration: config defaults disabled, shared thread/dedup types, async ancestry, owner-root signing, and semantic-index assembly. Check normal/build edges for absence of `heddle-repo`, `rusqlite`, and `libsqlite3-sys`; once the new Sley release is selected, also check absence of Ureq and the local worktree/hooks/sequencer engines. Run separately from the full workspace so feature unification cannot hide leaks.

Validation: the config build with defaults disabled passed locally at `5e3a699`, and its graph excludes `heddle-repo`/SQLite. It still includes Ureq and local Sley engines via `heddle-objects`. The subsequent `438454c` release changes leave these dependency/API gaps unchanged. The two publish-integrity checks now pass.

Weft will handle its own import/manifest cleanup, local test-adapter separation, and Reqwest adapter after these upstream boundaries land.

forge-zephyr:
The three extraction gaps from the earlier comment are addressed. The isolated hosted-consumer check passed locally and in CI at `0616f43`, excluding `heddle-repo`, SQLite, Ureq, and local Sley engines. The example import that failed the workspace build is also fixed in `c3eae03`.

While downstream adoption is waiting on the Sley release, please improve these remaining API boundaries:

1. **Require only read access for semantic assembly.** `SemanticIndexBuilder` currently requires the full writable `ObjectStore`, including for `build_root_deferred`, which returns generated blobs without persisting them. Base assembly on the existing `ObjectSource` interface and keep persistence in a separate helper or method with an additional write bound. Preserve parent-index reuse and all existing work/output limits. This lets Weft pass its staged, authorized source and control publication separately. Validate the deferred path with a source that implements only the read interface, and verify equivalent output to the existing persisted path.

2. **Return typed semantic failures.** Cancellation and exhausted source/output budgets should be distinguishable from malformed objects and underlying storage failures without matching `InvalidObject` message strings. Preserve enough information to distinguish deadline expiry from explicit cancellation where the budget API exposes it. Keep transport/status mapping in Weft; the shared library should expose semantic error variants. Add tests asserting those distinctions so Weft can map deadline, cancellation, and resource exhaustion correctly.

3. **Make async ancestry work bounded and test the extracted algorithm directly.** The extraction changes the previous implementation to a direct graph walk. Add caller-supplied work limits and a runtime-independent cancellation/deadline mechanism for hosted use. Limit/cancellation failures must remain distinct from `Ok(false)`; a completed traversal is needed to conclude non-ancestry. Document missing-parent behavior and add leaf-level tests for that behavior, source errors, merge histories, early termination, and exhaustion/cancellation. Retain parity coverage against the repository implementation for complete graphs. Avoid adding a Tokio dependency to the leaf API; callers remain responsible for cancellation/timeouts around blocked source I/O.

These improve the now-working dependency separation and make the shared APIs easier for Weft to consume. Keep the isolated consumer graph gate passing, rerun the workspace checks after these changes, and update the PR's testing section with the actual validation results.

